PR #1, opened 2020-06-26
Use pub tracks + Cleanup on pub disconnect

tarrencev:
Update to use pub tracks for subscribe (rather than passing them in). Properly cleanup pub ice disconnects too.

I removed the old examples and will open a PR with some new ones.


pkg/rtc/transport/webrtctransport.go
```diff
@@ -236,9 +236,10 @@ func NewWebRTCTransport(id string, options RTCOptions) *WebRTCTransport {
 	w.pc.OnICEConnectionStateChange(func(connectionState webrtc.ICEConnectionState) {
 		switch connectionState {
 		case webrtc.ICEConnectionStateDisconnected:
-			log.Errorf("webrtc ice disconnected")
+			log.Infof("webrtc ice disconnected for mid: %s", id)
 		case webrtc.ICEConnectionStateFailed:
-			log.Errorf("webrtc ice failed")
+		case webrtc.ICEConnectionStateClosed:
```
tarrencev, line 241: Should we handle `Closed` the same way as `Failed`? I actually haven't seen `Closed` triggered when doing `pc.Close()` with a pion client. Not sure if i am misunderstanding.

tarrencev:
Going to merge this and keep working on it. Feel free to leave feedback and I can address it in a follow up PR

PR #6, opened 2020-06-28
support trickle ice

tarrencev:
adds client trickle support. i don't think we need to handle trickle on the sfu side, since we wouldn't be interacting with stun/turn.

i think ideally we should have a `mid => pc` mapping, so we can handle trickle for a pc without having to know about routers. but i think that we can update that once we figure out how we want to multiplex streams/tracks


pkg/proto/sfu/sfu.proto
```diff
@@ -62,4 +63,12 @@ message PublishOptions {
...
     uint32 bandwidth = 1;
     bool transportcc = 2;
-}
\ No newline at end of file
+}
+
+message TrickleRequest {
+    string mid = 1;
+    string candidate = 2;
+    string pub = 3;
```
tarrencev, line 71: should be able to remove `pub` at some point and simplify things

cloudwebrtc:
@tarrencev Is there any way to broadcast trickle candidates from the SFU side to the client/browser?

tarrencev:
@cloudwebrtc it is possible but is it necessary? from my understanding that is only necessary if sfu is using stun/turn? not sure if i misunderstand

cloudwebrtc:
@tarrencev Normally, the turn/stun relay is not required, but in some extreme cases, such as ion-sfu is placed behind a firewall or NAT, you must use the turn proxy to forward from the edge server, or use stun.

tarrencev:
ok makes sense. i can add support for that too

tarrencev:
@cloudwebrtc updated to support trickle both ways

cloudwebrtc:
@tarrencev  LGTM

PR #9, opened 2020-06-29
simplify transport liveness management

tarrencev:
Working on simplifying the transport liveness management. Basically there are two ways a transport is closed. From gRPC by closing a stream or through ICE entering a `failed` or `closed` state. I changed the logic to be bottom-up, closing a transport bubbles up and deletes router state then rtc state.

For gRPC, since we are maintaining a streaming connection, as soon as the client closes the corresponding pub/sub will be closed.
For ICE, instead of tracking liveness internally, we can use ICE state. So we can remove liveness check in the router. For RTPTransport we can manage timeout internally.


pkg/rtc/rtc.go
```diff
@@ -160,11 +160,6 @@ func check() {
 		}
 
 		for id, router := range routers {
-			if !router.Alive() {
```
tarrencev, line 163: This is handled through ICE timeout now

jbrady42:
:+1:

PR #11, opened 2020-06-30
Extract publish/subscribe logic + update sdp parsing

tarrencev:
Update sdp parsing to use pion/sdp and add some testing


pkg/proto/sfu.proto
```diff
@@ -42,21 +42,15 @@ message SubscribeReply {
...
 
-message SessionDescription {
-    string type = 1; // "answer" | "offer" | "pranswer" | "rollback"
-    string sdp = 2;
-}
-
 message Options {
     uint32 bandwidth = 1;
     bool transportcc = 2;
-    string codec = 3;
```
tarrencev, line 61: Instead of explicitly providing a codec like this, we use the sdp codec ordering for priority. i'll provide a utility in the ion-sdk-js to specify preferred codec

PR #14, opened 2020-07-02
adding prometheus metrics server

leewardbound:
Does not include any SFU stats yet, but includes the prometheus metrics server launching with the `sfu` node.

+ I originally added counters to the `publish()` and `subscribe()` etc stream control methods, but we should actually get these for free when we add `go-grpc-prometheus`

+ I am working to add some `GetStats()` exposure, but (I'm not an expert, I might be wrong) -- I didn't like how GetStats always returns an increasing number for most metrics; they typically are more interesting as instantaneous values -- this is taking me some time to implement as I am a golang novice, and I think it's best if I commit this simple branch now so we can begin to use metrics

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -2,12 +2,15 @@
 package main
 
 import (
+	"net/http"
+
 	conf "github.com/pion/ion-sfu/pkg/conf"
 	"github.com/pion/ion-sfu/pkg/log"
 	sfu "github.com/pion/ion-sfu/pkg/node"
 	"github.com/pion/ion-sfu/pkg/rtc"
 	"github.com/pion/ion-sfu/pkg/rtc/plugins"
 	"github.com/pion/webrtc/v2"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 func init() {
@@ -34,6 +37,17 @@ func init() {
 
 	if err := rtc.InitRTP(conf.Rtp.Port, conf.Rtp.KcpKey, conf.Rtp.KcpSalt); err != nil {
 		panic(err)
+	}
+
+	if conf.Plugins.Metrics != "" {
+		go func() {
+			log.Infof("Serving metrics at %s/metrics", conf.Plugins.Metrics)
+			http.Handle("/metrics", promhttp.Handler())
+			err := http.ListenAndServe(conf.Plugins.Metrics, nil)
+			if err != nil {
+				panic(err)
+			}
+		}()
 	}
 
 	pluginConfig := plugins.Config{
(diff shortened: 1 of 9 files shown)


tarrencev:
For simplicity, I don't think we should introduce prometheus in this repo. `ion` reimplements `cmd` in at https://github.com/pion/ion/pull/309/files#diff-5b63bcde9120c2b022b48409414a664bR67 so perhaps it makes more sense there (once we merge the migration). We can add some generalized stats endpoints here, but how they are aggregated/displayed should be up to the consumer

leewardbound:
I understand the instinct to maintain simplicity, and at first glance, I recognize that this seems to add little extra functionality beyond just consuming a TCP port. Like I mentioned in our chat, there's actually a few unique (cant get them elsewhere) metrics that are exposed even by just this simple change, like `go_goroutines` and some garbage collection stats. My original goal was to include in this PR figures from `PeerConnection.GetStats` API, specifically metrics for bandwidth consumption (per connection, tagged with metadata) and UDP ports (one of our tightest resources, until the proposed PC changes are implemented). This needs to run by iterating all the PCs and accumulating the stats, which can only execute on the SFU.

I haven't finished my new CountRouterMetrics method yet, but in my `feature/prometheus-advanced` branch we count all the metrics in the `rtc/webrtctransport.go` file (by adding a `CountMetrics(cycleSeconds int)` method to the Transport interface). Just adding the server line in `cmd` on the `ion` repo will not suffice to expose these stats (and I also have a really hard time with the idea of having 2 "builds" of SFU, one from `ion` that includes metrics in the build and one from `ion-sfu` that does not, honestly I'd sooner scrap telemetry entirely than suggest that).

Telemetry we can get without introspecting `GetStats` API is limited to "per-container" aggregations; I think it's common for a piece of core infrastructure like SFU to expose a metrics endpoint to enable operational graphs like "bytes per second, by codec" or of open UDP ports (a scarce resource that is not typically exposed by `cAdvisor`). With the new proposed UDP port changes, you could make the argument this becomes less operationally critical.

Ultimately, I think we'll come back and add it some day, but this is A-OK and I fully own that the utility of "just exposing it" without the extra metrics isn't useful here - If you don't care to expose at least the metrics from `go-grpc-prometheus` like you showed me last night (I thought you were excited about that :) ) then we don't need this!

tarrencev:
i for sure want the prometheus logging! just think it belongs in `ion` rather than here. since `ion` already reimplements `cmd` for sfu (to support service discovery). lets work on exposing the `GetStats` api. I'm happy to help with implementation

PR #28, opened 2020-07-13
Dynamic track handling + payload type matching

### LINKED ISSUE #23: iOS/Apple, Surface Pro - not able to receive the stream (the root cause provided)
### Your environment.
  - Browser: Safari 13.1.2 with Chrome on Windows 
  - Other Information - see below

### What did you do?
1. Opened Chrome on Windows and started the stream.
2. Opened Safari or Chrome on macOS and couldn't receive the stream.

### What did you expect?
See the stream.

### What happened?
I couldn't see it.

After https://github.com/pion/ion-sdk-js/pull/28 fix, I was able to experiment a bit more with Safari.

It turns out, that in the SDP Offer Safari has:
image

(...)
m=video 9 UDP/TLS/RTP/SAVPF 96 97 98 99 100 101 127 125 104
(...)
a=rtpmap:96 H264/90000
(...)

but in the SDP answer we have:
m=video 9 UDP/TLS/RTP/SAVPF 96
(...)
a=rtpmap:96 VP8/90000
--------

For Safari, to get VP8/90000, the SAVPF should be set to 100. 

96 works flawlessly with Chrome (on other OS than Mac/iOS), as with Chrome VP8/90000, indeed, it is 96.

To fix that, for maximum compatibility in the future, we probably need to look-up VP8 in the SDP Offer and return the same value to the browser.

Potential solution/fix/help here: https://github.com/pion/webrtc/issues/716#issuecomment-502609027

tarrencev:
This is a fairly significant refactor of how we handle tracks + payload types. It updates track handling to be dynamic, so it isn't necessary to know about tracks before subscribing to a pc. When a track is added/removed from a pub, the sub will be notified. This is necessary for handling renegotiation.

It also moves away from mapping codecs based on payload type and instead uses codec names. This is based on a conversation in #pion which mentioned that payload types are dynamic and codec types shouldn't be assumed from them. This should resolve #23.


pkg/media/engine.go
```diff
@@ -0,0 +1,112 @@
...
+
+const (
+	mediaNameAudio    = "audio"
+	mediaNameVideo    = "video"
+	iOSH264Fmtp       = "level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42e01f"
+	firefoxH264Fmtp97 = "profile-level-id=42e01f;level-asymmetry-allowed=1"
+)
+
+var (
+	rtcpfb = []webrtc.RTCPFeedback{
```
tarrencev, line 20: Should set these based on SDP

pkg/media/engine.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/pion/sdp/v2"
+	"github.com/pion/webrtc/v2"
+)
+
+const (
+	mediaNameAudio = "audio"
+	mediaNameVideo = "video"
+	// iOSH264Fmtp       = "level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42e01f"
+	// firefoxH264Fmtp97 = "profile-level-id=42e01f;level-asymmetry-allowed=1"
```
tarrencev, line 16: I'm not sure if we need these anymore. Since we are using `fmtp` specified in the offer sdp

sibis:
Great work @tarrencev 👏🏽

PR #45, opened 2020-07-20
Single PC

tarrencev:
This is pretty much a rewrite of `ion-sfu` to support single peer connection between a client and sfu with all pub/sub tracks on the same transport. With this, everything is handled at the track level. Each receiver track gets it's own router. Peer connections belong to a room(s) and are automatically subscribed to any other peers in the same room.

The endpoints are basically:
1) Join
2) Offer
3) Answer
4) Trickle

The benefit of this approach:
1) It is more robust since we only need a single ice connection. Either everything fails or everything succeeds. No intermediate states where a some users are connected, others aren't
2) Much less chatter. Just a single ice exchange per user per sfu. Faster connects for new participants too.
3) Less port usage (n instead of n^2)

This is a rough wip implementation. I removed a bunch of stuff to simplify things (like rtp transport + forwarder) that we can add back. Interested in any initial feedback!


pkg/receiver.go
```diff
@@ -0,0 +1,265 @@
...
+	pliCycle     int
+	rembCycle    int
+	maxBandwidth int
+}
+
+// VideoReceiverConfig .
+type VideoReceiverConfig struct {
+	TCCOn         bool `mapstructure:"tccon"`
+	REMBCycle     int  `mapstructure:"rembcycle"`
+	PLICycle      int  `mapstructure:"rembcycle"`
```
adwpc, line 98: should be `mapstructure:"plicycle"`

pkg/receiver.go
```diff
@@ -0,0 +1,265 @@
...
+type Receiver interface {
+	Track() *webrtc.Track
+	GetPacket(sn uint16) *rtp.Packet
+	ReadRTP() (*rtp.Packet, error)
+	ReadRTCP() (rtcp.Packet, error)
+	WriteRTCP(rtcp.Packet) error
+	Close()
+}
+
+// AudioReceiver receives a video track
```
adwpc, line 29: receives a audio track

PR #46, opened 2020-07-26
add wss support and readme

adwpc:
#### Description
add wss support and readme

#### Reference issue
allow test online/LAN


tarrencev:
What do you think about rather including a simple way to standup a Caddy reverse proxy that handles ssl termination?

joe-getcouragenow:
> What do you think about rather including a simple way to standup a Caddy reverse proxy that handles ssl termination?

You can also do it without Caddy too if you want using the same SSL lib (github.com/caddyserver/certmagic)  as Caddy

Here is a simple example
https://github.com/lithdew/flatend/blob/master/cmd/flatend/main.go#L153

For many servers you need to use ETCD as the storage place, so that when new servers boot they dont try to make another Cert, but instead check to get the one fro ETCD.

adwpc:
https://github.com/pion/ion-sfu/pull/55
new pr

adwpc:
> > What do you think about rather including a simple way to standup a Caddy reverse proxy that handles ssl termination?
> 
> Hey @tarrencev
> You can also do it without Caddy too if you want using the same SSL lib (github.com/caddyserver/certmagic) as Caddy
> 
> Here is a simple example
> https://github.com/lithdew/flatend/blob/master/cmd/flatend/main.go#L153
> 
> For many servers you need to use ETCD as the storage place, so that when new servers boot they dont try to make another Cert, but instead check to get the one fro ETCD.


tryed but not work. don't know how to use with ws.

adwpc:
> What do you think about rather including a simple way to standup a Caddy reverse proxy that handles ssl termination?

not family with canddy. : (
i think should support config cert files by hand.

joe-getcouragenow:
> > > What do you think about rather including a simple way to standup a Caddy reverse proxy that handles ssl termination?
> > 
> > 
> > Hey @tarrencev
> > You can also do it without Caddy too if you want using the same SSL lib (github.com/caddyserver/certmagic) as Caddy
> > Here is a simple example
> > https://github.com/lithdew/flatend/blob/master/cmd/flatend/main.go#L153
> > For many servers you need to use ETCD as the storage place, so that when new servers boot they dont try to make another Cert, but instead check to get the one fro ETCD.
> 
> tryed but not work. don't know how to use with ws.

@adwpc If you have a PR i can look at it.

the other way is to just gen the certs using a bash script onto disk and then your good to go.
https://github.com/acmesh-official/acme.sh
- works for me

mkcert ( for local with real HTTPS in the desktop browser and mobile)
- helps local testing.

tarrencev:
@joe-getcouragenow you can see the pr here #55. would be great if you could improve the flow!

PR #55, opened 2020-08-01
Json rpc wss1

adwpc:
#### Reference issue
Fixes #...


tarrencev:
@adwpc i removed the `.pem` files and moved the `https` logic to the `json-rpc` readme. I would prefer to keep the example as simple as possible

PR #61, opened 2020-08-02
fix grpc example and doc

### LINKED ISSUE #56: fix examples
`pub-from-browser`, `pub-from-disk`, and `sub-to-browser` gRPC examples need to be updated to the new interface

adwpc:
#### Reference issue
Fixes #56


adwpc:
lost pub-from-disk example

adwpc:
done

PR #65, opened 2020-08-02
Point Slack badge to Pion Slack info page

vicramon:
The current link doesn't include the invite link or any info on how to get to it. 

The updated one is the same link that's used on the main pion/ion repo.

Might save the next person 10 minutes :)


Sean-Der:
:fire: thanks @vicramon

PR #70, opened 2020-08-04
relay transport

tarrencev:
This PR implements basic sfu relay support. It introduces a relay packet which contains some metadata for the track being sent.

The relay packet looks like this right now and is prepended to every rtp/rtcp packet that is relayed (the reserved part is for including other metadata in the future). Track payload types are normalized before being relayed so we can make assumptions without passing SDP info.

```
/*
 *  0                   1                   2                   3
 *  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 * | V |                       reserved                            |
 * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 * |                         session id                            |
 * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 */
```

The sfu creates a relay Server on init and listens there for new relay connections. When it receives a new connection, it demuxes based on `session id` then `rcp/rtcp`. So a muxed track data can be sent over a single connection.

When a new session id is received, a `RelayTransport` is constructed. The `RelayTransport` listens for an `RTP` stream to arrive and constructs a Track + Receiver to handle it.


adwpc:
1. Maybe we need a simple auth
2. add a length: length of relay packet,  two packet may merged together?
3. peer id? may peers can join one session
4. relay infos:from srcip to dstip
5. track infos: audio/video pt ssrc  simulcat info

tarrencev:
1. For auth, i think kcp/quic would be correct approach? RTP over UDP is just simplest first iteration. or do you think we need something additional?
2. Good idea. I think we will need some sequence number then too? to locate parts?
3. I dont think peer id is necessary. Signaling layer should track MediaStream.id/ssrc to peer id globally so can associate later
4. Not familiar with this? 
5. pt/ssrc is in rtp header. do we need additional simulcast info?
6. I dont think we want app level info in relay header

It seems we need a way to synchronize media info between nodes. I dont think it makes sense to send it as a header to every packet. maybe we want an sdp exchange?

PR #73, opened 2020-08-05
Add test for peer.OnClose() listener.

kevmo314:
For context, I'm debugging an issue where `peer.Close()` seems to hang.

Apparently it's not from just calling `.Close()`, but here's a free test. :)


tarrencev:
awesome thanks!

PR #77, opened 2020-08-06
Send ICECandidate using ToJSON()

cameronej:
Send ICECandidate using `ToJSON()` rather than `String()`, same as in the grpc server. 
`ToJSON()` returns `sdpMLineIndex`  along with the `candidate` string, where as `String()` only returns the `candidate` string.


tarrencev:
thanks!

cameronej:
No worries

PR #79, opened 2020-08-07
Expose OnTrack and OnConnectionStateChange handlers.

kevmo314:
These are necessary so the biz node gets a signal when a new track is emitted to associate it with a uid and so the biz node can disconnect clients that end up in a terminal PeerConnectionState automatically.

### DIFF
--- a/pkg/webrtctransport.go
+++ b/pkg/webrtctransport.go
@@ -170,6 +170,16 @@ func (p *WebRTCTransport) OnNegotiationNeeded(f func()) {
 	}
 }
 
+// OnTrack handler
+func (p *WebRTCTransport) OnTrack(f func(*webrtc.Track, *webrtc.RTPReceiver)) {
+	p.pc.OnTrack(f)
+}
+
+// OnConnectionStateChange handler
+func (p *WebRTCTransport) OnConnectionStateChange(f func(webrtc.PeerConnectionState)) {
+	p.pc.OnConnectionStateChange(f)
+}
+
 // NewSender for peer
 func (p *WebRTCTransport) NewSender(intrack Track) (Sender, error) {
 	to := p.me.GetCodecsByName(intrack.Codec().Name)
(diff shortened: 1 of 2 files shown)


kevmo314:
Oh hrm, `OnTrack` is going to have problems because only one event handler can be assigned. @tarrencev do you have any thoughts for how to associate new tracks with a user if they don't involve an `OnNegotiationNeeded` SDP exchange? The biz/islb node will never hear from the sfu node because it doesn't have a way of hooking into the existing `PeerConnection`.

kevmo314:
Actually, nevermind just adding another handler layer will work.

tarrencev:
Looks good. Would you mind just adding basic tests for this? You will prob need to use the mutex for since the handlers are called in a separate goroutine

kevmo314:
Done, added tests and mutex guards. :)

pkg/webrtctransport.go
```diff
@@ -159,17 +164,35 @@ func (p *WebRTCTransport) AddICECandidate(candidate webrtc.ICECandidateInit) err
...
+// OnTrack handler
+func (p *WebRTCTransport) OnTrack(f func(*webrtc.Track, *webrtc.RTPReceiver)) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.onTrackHandler = f
+}
+
+// OnConnectionStateChange handler
+func (p *WebRTCTransport) OnConnectionStateChange(f func(webrtc.PeerConnectionState)) {
+	p.mu.Lock()
```
tarrencev, line 191: Do we need to guard here? Since it is just calling to pc

pkg/webrtctransport.go
```diff
@@ -159,17 +164,35 @@ func (p *WebRTCTransport) AddICECandidate(candidate webrtc.ICECandidateInit) err
 
 // OnICECandidate handler
 func (p *WebRTCTransport) OnICECandidate(f func(c *webrtc.ICECandidate)) {
+	p.mu.Lock()
```
tarrencev, line 167: Do we need to guard here? Since it is just calling to pc

pkg/webrtctransport.go
kevmo314: Removed.

pkg/webrtctransport.go
kevmo314: Removed.

PR #80, opened 2020-08-09
Don't add sender to router if error

cameronej:
If there's an error when creating a new sender `WebRTCTransport.NewSender` the server will crash with `panic: runtime error: invalid memory address or nil pointer dereference` at `router.go` line 110 as the sender is nil. 

I added a `continue` to skip the nil sender, alternatively we could return an error from the function to not allow the user to join the session at all.

This issue occurred when a second user was joining the session, but did not have a webcam (so the SDP only had an audio line). The `NewSender` function returned the `errPtNotSupported` error as it was trying to add a sender for a video track, but no video codecs were in the `MediaEngine.codecs`. 

I think it would still be useful to just skip this sender (rather than not allow them to join the session) so the user can still participate in the session with audio only. 

To prevent this from the frontend you can add a video transceiver manually `pc.addTransceiver('video')` if the user hasn't added any video tracks.


tarrencev:
awesome thanks! i think this is the right solution. if a user mutes audio/video we should remove the track. would you mind adding a simple test for this flow?

cameronej:
No worries, I will add the test

tarrencev:
awesome!

PR #81, opened 2020-08-10
media engine updates

tarrencev:
if the client pc doesn't define codecs, use defaults. this should resolve the issue where the initial offerer doesn't have audio/video track and as such media engine doesn't contain codecs


tarrencev:
@cameronej i think this resolves the issue you were seeing previously. the offerer should still receive video/audio even if the initial offer doesn't contain audio/video tracks

cameronej:
Hi @tarrencev I've tested these changes and it does solve the issue I was having before (and does so without having to manually add the video transceiver on the client who has audio only)

PR #90, opened 2020-08-18
fix demos can not display local sdp.

cgojin:
fix demos can not display local sdp.
fix base64 decode panic: illegal base64 data at input byte xxx.

### DIFF
--- a/examples/pub-from-browser/jsfiddle/demo.js
+++ b/examples/pub-from-browser/jsfiddle/demo.js
@@ -27,7 +27,7 @@ navigator.mediaDevices.getUserMedia({ video: true, audio: true })
 
 pc.oniceconnectionstatechange = e => log(pc.iceConnectionState)
 pc.onicecandidate = event => {
-  if (event.candidate === null) {
+  if (event.candidate !== null) {
     document.getElementById('localSessionDescription').value = btoa(JSON.stringify(pc.localDescription))
   }
 }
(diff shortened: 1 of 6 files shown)


examples/pub-from-browser/jsfiddle/demo.js
```diff
@@ -27,7 +27,7 @@ navigator.mediaDevices.getUserMedia({ video: true, audio: true })
 
 pc.oniceconnectionstatechange = e => log(pc.iceConnectionState)
 pc.onicecandidate = event => {
-  if (event.candidate === null) {
+  if (event.candidate !== null) {
```
tarrencev, line 30: it is necessary to wait for null here in order for ice to complete

cgojin: can wait for null in edge,  but can not wait for null in chrome/firfox.

tarrencev: making this change will create the offer before gathering is complete and the sdp exchange will fail. it will also create an offer for every trickle ice candidate. 

you can see the same is done here https://github.com/pion/webrtc/blob/master/examples/broadcast/jsfiddle/demo.js#L16

cgojin: You are right, chrome/firefox can wait for null, but it may take more than 30 seconds, maybe we can add a delay of 1 second to display the latest local sdp?

cgojin:
Resubmit after revision.

PR #91, opened 2020-08-19
fix base64 decode panic; quickly display sdp.

cgojin:
fix base64 decode panic: illegal base64 data at input byte xxx.
quickly display sdp in chrome/firefox.

### DIFF
--- a/examples/internal/signal/signal.go
+++ b/examples/internal/signal/signal.go
@@ -60,7 +60,7 @@ func Encode(obj interface{}) string {
 // Decode decodes the input from base64
 // It can optionally unzip the input after decoding
 func Decode(in string, obj interface{}) {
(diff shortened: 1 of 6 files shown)


tarrencev:
@cgojin i still dont think this `onicecandidate` changes are necessary. i would like to keep examples as simple as possible. it is ok if they need to wait 30s for gathering to complete. the right solution is to add trickle ice support to the examples but it has been left out for simplicity. im happy to add info the readme that is might require some time for gathering to complete.

re: the base64 change can you share an example that causes the panic?

cgojin:
this is ok.

I tested again many times, base64 decode no problem,  i double-clicked the text box and did not fully select all base64 string, the '=' at the end may not be selected.

PR #99, opened 2020-08-31
enhancement/new-transport-media-engine

decanus:
#### Description

Changes the `NewWebRTCTransport` function to use the `MediaEngine` rather than an `SessionDescription`, this enables the SFU to make offers.

### DIFF
--- a/pkg/session_test.go
+++ b/pkg/session_test.go
@@ -67,7 +67,12 @@ func TestSession(t *testing.T) {
 	<-gatherComplete
 
 	session := NewSession("session")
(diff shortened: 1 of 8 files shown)


pkg/session_test.go
```diff
@@ -359,8 +364,11 @@ func Test3PeerStaggerJoin(t *testing.T) {
 	assert.NoError(t, err)
 	err = remoteB.SetLocalDescription(offer)
 	assert.NoError(t, err)
+	engine := MediaEngine{}
```
decanus, line 367: @tarrencev I think this entire test will require refactoring.

PR #112, opened 2020-09-04
ci(workflow): switch to github workflow

tarrencev:
moving lint to github actions

### DIFF
--- a/.github/workflows/build-grpc.yaml
+++ b/.github/workflows/build-grpc.yaml
@@ -7,18 +7,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/setup-go@v1
-        with:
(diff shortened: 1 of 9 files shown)


.github/workflows/build-grpc.yaml
```diff
@@ -0,0 +1,24 @@
...
+          go-version: '1.15'
+
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~/.cache
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: ${{ runner.os }}-go-
+
```
steebchen, line 21:
```suggestion
```

This is not needed because it doesn't use private modules

PR #119, opened 2020-09-06
make recv-side-twcc work

### LINKED ISSUE #117: TWCC support
- [x] Recv-Side TWCC

adwpc:
#### Description
make recv-side-twcc work 

#### Reference issue
Fixes #117

### DIFF
--- a/pkg/mediaengine.go
+++ b/pkg/mediaengine.go
@@ -16,11 +16,9 @@ const (
 
 var (
 	rtcpfb = []webrtc.RTCPFeedback{
-		{Type: webrtc.TypeRTCPFBGoogREMB},
 		{Type: webrtc.TypeRTCPFBCCM},
 		{Type: webrtc.TypeRTCPFBNACK},
 		{Type: "nack pli"},
-		// {Type: webrtc.TypeRTCPFBTransportCC},
 	}
 )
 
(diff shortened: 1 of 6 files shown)


tarrencev:
will take a look tomorrow!

pkg/mediaengine.go
```diff
@@ -16,11 +16,9 @@ const (
 
 var (
 	rtcpfb = []webrtc.RTCPFeedback{
-		{Type: webrtc.TypeRTCPFBGoogREMB},
 		{Type: webrtc.TypeRTCPFBCCM},
 		{Type: webrtc.TypeRTCPFBNACK},
 		{Type: "nack pli"},
-		// {Type: webrtc.TypeRTCPFBTransportCC},
```
tarrencev, line 23: We dont need to specify either of these?

pkg/receiver.go
```diff
@@ -225,7 +225,7 @@ func (v *WebRTCVideoReceiver) receiveRTP() {
 
 		v.buffer.Push(pkt)
 
-		if v.feedback == webrtc.TypeRTCPFBTransportCC {
```
tarrencev, line 228: shouldn't this be set based on whether client supports it?

adwpc: Yes.
Setting some params by JS-SDK is better. We can use it by settting on sfu config file recently.
I think we can merge first and do a better api setting on sdk when JS-SDK is ready

PR #121, opened 2020-09-07
Doc and example bug fixes

TannerGabriel:
#### Description
This pull request addresses the following issues:

- The docker commands in the `README.md` file reference the wrong Dockerhub user `pion` instead of `pionwebrtc`. Therefore when you try to pull the image, you get the following error message: `docker: Error response from daemon: pull access denied for pion/ion-sfu, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.`.
- Typo in the `pub-from-browser` where the wrong name is used to reference the div. As a result the application throws an error when clicking on the `Start session` button.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ go build ./cmd/server/json-rpc/main.go && ./main -c config.toml
 If you prefer a containerized environment, you can use the included Docker image
 
 ```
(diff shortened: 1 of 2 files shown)


tarrencev:
awesome thanks!

PR #123, opened 2020-09-08
add pub from chrome example

jinleileiking:
This time, you do not need cmdline, no jsfiddle,  just a chrome browser,  open & click!


tarrencev:
awesome! i think we can just replace `pub-from-browser` with this!

would you mind doing that and putting `tool` into `examples/internal/crypto.go`

examples/pub-from-chrome/sfu.js
```diff
@@ -0,0 +1,207 @@
...
+//         // }]
+//     })
+
+
+//     // 1. send subscribe  sdp
+//     pcSubcribe.onicecandidate = event => {
+//         if (event.candidate === null) {
+//             log("SDP chrome ->  sfu:\n" + pcSubcribe.localDescription.sdp)
+//             var sendData = {type:'subscribe', sdp:pcSubcribe.localDescription.sdp, name:name}
+//             sock.send(JSON.stringify(sendData));
```
TannerGabriel, line 179: Typo: Should be `socket.send()` instead.

examples/pub-from-chrome/sfu.js
```diff
@@ -0,0 +1,207 @@
...
+//     pcSubcribe.addTransceiver('audio', {'direction': 'recvonly'})
+//     pcSubcribe.addTransceiver('video', {'direction': 'recvonly'})
+
+//     pcSubcribe.createOffer()
+//         .then(d => pcSubcribe.setLocalDescription(d))
+//         .catch(log)
+
+//     // 4. receive data
+//     pcSubcribe.ontrack = function (event) {
+//         var el = document.getElementById('remote')
```
TannerGabriel, line 192: The `remote` HTML element is currently a div, but you use it like it is a video element. I think you meant to create a video tag inside the div and display the video using this tag.

TannerGabriel (changes requested):
Great project! Here are some small things I found that may be valuable to correct.

jinleileiking:
I will remove the subscribe code.

examples/pub-from-chrome/sfu.js
jinleileiking: done . delete unused code.

jinleileiking:
@tarrencev @TannerGabriel  ready for merge. Give a look when free. 

I do not overwrite pub-from-browser because I do not know whether the writer of `pub-from-browser` is happy for rewriting.

examples/pub-from-chrome/index.html
```diff
@@ -0,0 +1,22 @@
...
+    </head>
+
+    <body>
+        remote</br>
+        <div id="remote" width="320" height="240" ></div>
+        local</br>
+        <div id="local" width="160" height="120" ></div> <br />
+
+        <button onclick="window.Pub('pub')">Publish</button>
+        <button onclick="window.Sub('sub')">Subscribe</button>
```
TannerGabriel, line 18: The subscribe button is never used since you deleted the subscribe feature.

TannerGabriel (changes requested):
Great changes the application works now! Just one minor thing that should be addressed because it could confuse some users.

tarrencev:
Thanks for the great example! I migrated this to be a custom signaling example

PR #126, opened 2020-09-08
Detect disabled tracks (#122)

### LINKED ISSUE #122: MediaStreamTrack.enabled = false, NACKS
### Your environment.
  - Version: *LATEST*
  - Browser: *LATEST*
### What did you do?

When video track is disabled, the SFU starts forwarding NACKS continuously.

### What did you expect?

No NACKS messages.

### What happened?

Maybe it's possible to add a parameter to let know pion that a track it's going to be disabled, or limit rate the NACKS

OrlandoCo:
#### Description
This PR it's an attempt to detect a disabled track using the payload size.

#### Reference issue
Fixes #122

### DIFF
--- a/pkg/buffer.go
+++ b/pkg/buffer.go
@@ -24,6 +24,10 @@ const (
 
 	// default buffer time by ms
 	defaultBufferTime = 1000
+
+	// Threshold to detect that current packet contains a full black pixel frame by payload size from disabled video
+	// track
+	blackFrameThreshold = 450
 )
 
 func tsDelta(x, y uint32) uint32 {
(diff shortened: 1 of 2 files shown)


tarrencev:
@OrlandoCo thanks for raising this! Would you mind moving it to a branch on `pion` so it is easier for us to collaborate. I've added you as a member. Also, it would be great to look into how jitsi, ect handle this

OrlandoCo:
Investigating more, I found that the root of the NACKS was because packets were being cleaned before they could be read, the reason is that when video is muted, the packages per second drops to 15.

The quickest solution is to increase the buffertime, and the nacks are gone. I will to investigate better the best defaylt buffertime vaule to prevent this problem.

Issue #101, opened 2020-09-01
feat(relay): add example


tarrencev:
`ion-sfu` supports sfu <> sfu relay. we should add an example.

we can add a gRPC endpoint that can be called to connect to a remote sfu + session which will automatically start relaying the tracks. `ion-avp` does something similar already.

obasajujoshua31:
Implementation plan 


1. update sfu.proto with new rpc
rpc ConnectRemote(stream Signalrequest) returns (stream signalResponse) {}

2. compile the sfu.proto that will generate go stubs and add connectRemote server and client implementation

3. in connectRemote server, I will call Server.signal with the stream coming from connectRemote stream

4. Add examples in the examples directory.

tarrencev:
Awesome! I think this is a solid direction

For the api, I think we can do something like:
```
rpc Relay(RelayRequest) returns (RelayResponse) {}

message Request {
    string sfu = 1; 
    string sid = 2;
}
```

Calling `Relay` will create an `SFUClient` for the given sfu address if one doesn't exist. Then `SFUClient` can join the remote `SFUServer` and begin signaling.

PR #130, opened 2020-09-09
Feat: relay-101

obasajujoshua31:
added relay endpoint to sfu.proto

ran ./build_protos.sh

added server implementation to relay

added examples for relay in examples

#### Reference issue
https://github.com/pion/ion-sfu/issues/101
Fixes #...


cmd/server/grpc/main.go
```diff
@@ -38,11 +39,18 @@ var (
 	errNoPeer = errors.New("no peer exists")
 )
 
-type server struct {
-	pb.UnimplementedSFUServer
+// Server struct
+type Server struct {
```
tarrencev, line 43: can we keep this private?

do we no longer need `pb.UnimplementedSFUServer`?

are you sure we need to change this?

cmd/server/grpc/main.go
```diff
@@ -110,15 +118,19 @@ func main() {
...
 	lis, err := net.Listen("tcp", addr)
 	if err != nil {
 		log.Panicf("failed to listen: %v", err)
 	}
 	log.Infof("SFU Listening at %s", addr)
 	s := grpc.NewServer()
-	pb.RegisterSFUServer(s, &server{
-		sfu: sfu.NewSFU(conf.Config),
+	sfuServer := NewSFUServer(conf.Config)
+	pb.RegisterSFUService(s, &pb.SFUService{
```
tarrencev, line 131: I feel like we didn't need to change this?

I think it can be like before:
```
pb.RegisterSFUServer(s, &server{
    sfu: sfu.NewSFU(conf.Config),
})
```

and revert changes above too (other than line 121)

cmd/server/grpc/main.go
```diff
@@ -351,3 +365,56 @@ func (s *server) Signal(stream pb.SFU_SignalServer) error {
...
+// Relay endpoint implementation
+func (s *Server) Relay(ctx context.Context, request *pb.RelayRequest) (*pb.RelayResponse, error) {
+
+	conn, err := grpc.Dial(request.Sfu, grpc.WithInsecure(), grpc.WithBlock())
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "%s", err)
+	}
+	defer conn.Close()
+
+	sfuClient := pb.NewSFUClient(conn)
```
tarrencev, line 378: lets cache this in a map `server`. we can use `request.Sfu` as key. if it already exists, we dont need to create a new client

cmd/server/grpc/main.go
```diff
@@ -351,3 +365,56 @@ func (s *server) Signal(stream pb.SFU_SignalServer) error {
...
+
+	client, err := sfuClient.Signal(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "%s", err)
+	}
+
+	me := sfu.MediaEngine{}
+	me.RegisterDefaultCodecs()
+
+	t, err := sfu.NewWebRTCTransport(ctx, &sfu.Session{}, me, sfu.WebRTCTransportConfig{})
```
tarrencev, line 388: We should use the session if it exists on the sfu, otherwise create a new session with `sid`

cmd/server/grpc/main.go
```diff
@@ -351,3 +365,56 @@ func (s *server) Signal(stream pb.SFU_SignalServer) error {
...
+		Payload: &pb.SignalRequest_Join{
+			Join: &pb.JoinRequest{
+				Sid: request.Sid,
+				Offer: &pb.SessionDescription{
+					Type: offer.Type.String(),
+					Sdp:  []byte(offer.SDP),
+				},
+			},
+		},
+	})
```
tarrencev, line 413: We will need to create a goroutine here to handle renegotiation + trickle ice. But lets address issues above first

cmd/server/grpc/main.go
obasajujoshua31: when I compiled the .proto, the line was removed.

obasajujoshua31: golang lint was making noise because of the server struct being private

cmd/server/grpc/main.go
obasajujoshua31: my compile is version 7, i guess, it changed things

cmd/server/grpc/main.go
obasajujoshua31: okay.

adwpc:
shoud restart after simulcast stable?

PR #140, opened 2020-09-10
Move cache cleaner after nack pair calculation

### LINKED ISSUE #122: MediaStreamTrack.enabled = false, NACKS

OrlandoCo:
This changes helps to prevent cleaning of packets before being read in the nack pair calculation, this may happen when a video is muted and slow connections.

#### Reference issue
Fixes #122

### DIFF
--- a/pkg/buffer.go
+++ b/pkg/buffer.go
@@ -120,17 +120,13 @@ func (b *Buffer) Push(p *rtp.Packet) {
 	b.pktBuffer[p.SequenceNumber] = p
 	b.lastPushSN = p.SequenceNumber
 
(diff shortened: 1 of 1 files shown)


tarrencev:
awesome! thanks for the pr

PR #145, opened 2020-09-11
fix(test): remove track race condition

tarrencev:
is that what you were thinking?


OrlandoCo (approved):
Yes with that delay the test should work now 👍

PR #152, opened 2020-09-12
cut sdp from sfu side

adwpc:
#### Description
1、worked for json-rpc
2、not wored for grpc, there are some error about protocol struct gened by protoc
3、Maybe cut from browsor is better
4、It cuted almost 80 lines per sdp about chrome :)

#### Reference issue
Fixes #...


tarrencev:
nice! as we discussed in slack, i think it is better to signal codec preferences in sdp. so modify sdp on client side. that way we can keep the rpc interface simple and signal in a unified way

adwpc:
Agree.

PR #158, opened 2020-09-13
chore(ctx): safe context done on channel writers

OrlandoCo:
#### Description

Since select statements can proceed in a pseudo-random selection, it's possible that runtime choose another case even if ctx.Done is called.

It's good practice to double check for context done that are in select, to prevent writing on closed channels.


tarrencev:
interesting. can we remove some of the selects then?

OrlandoCo:
> interesting. can we remove some of the selects then?

Yes we can clean some selects, I will push new changes.

OrlandoCo:
@tarrencev Since closing a channel is not mandatory, I think it's safer to not close the channel from a reader when are many go routines writing to it without some kind of sync like waitgroup.

tarrencev (approved):

PR #159, opened 2020-09-13
fix(sender): delay sending rtp to avoid chrome issue 10139

tarrencev:
@OrlandoCo given that this is a chrome issue, i decided to take the simple naive solution you suggested for now. hopefully this solves it so we dont need to add a complex solution and maybe chrome will fix it soon :)

### DIFF
--- a/pkg/sender.go
+++ b/pkg/sender.go
@@ -69,6 +69,12 @@ func NewWebRTCSender(ctx context.Context, track *webrtc.Track, sender *webrtc.RT
 }
 
 func (s *WebRTCSender) sendRTP() {
+	// There exists a bug in chrome where setLocalDescription
+	// fails if track RTP arrives before the sfu offer is set.
+	// We deplay sending RTP here to avoid the issue.
+	// https://bugs.chromium.org/p/webrtc/issues/detail?id=10139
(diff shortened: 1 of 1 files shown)


pkg/sender.go
```diff
@@ -69,6 +69,12 @@ func NewWebRTCSender(ctx context.Context, track *webrtc.Track, sender *webrtc.RT
 }
 
 func (s *WebRTCSender) sendRTP() {
+	// There exists a bug in chrome where setLocalDescription
+	// fails if track RTP arrives before the sfu offer is set.
+	// We deplay sending RTP here to avoid the issue.
+	// https://bugs.chromium.org/p/webrtc/issues/detail?id=10139
+	time.Sleep(500 * time.Millisecond)
```
tarrencev, line 76: I think this should be long enough 🤷‍♂️

OrlandoCo: I don't have encountered this bug before (and the SFU test server is in another country), so I think that this delay it's long enough to be sure that all the signaling is complete, and don't feel any unnecessary delay in the stream.

OrlandoCo:
Looks good, simple it's always better 🙌

OrlandoCo (approved):

PR #169, opened 2020-09-15
Update  pub-sub-in-browser example readme

Lumexralph:
#### Description

The command line flag `-p` has been changed to `-a`. The current startup command in the readme will lead to an error indicating that the flag is not valid.


tarrencev:
thanks!

PR #170, opened 2020-09-15
Broadcast channel & Session meta-data

billylindeman:
#### Description
- Adds the ability to broadcast messages to users in a session
- Adds metadata storage for tracks to session (and broadcasts updates)

These additions are useful for attaching user-id's and type information to streams in a session

### DIFF
--- a/cmd/server/json-rpc/main.go
+++ b/cmd/server/json-rpc/main.go
@@ -11,13 +11,12 @@ import (
 	"os"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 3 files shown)


tarrencev:
@billylindeman thank you for this proposal. This is outside of the scope of `ion-sfu` though. It should be implemented at the signaling layer or `ion-sfu` can be extended within your particular application. The `json-rpc` signaling included is there mostly as an example or to support simple use cases.

PR #173, opened 2020-09-15
fix(config): Fix config naming

OrlandoCo:
#### Description
Fix a bad config name introduced in last commit


tarrencev (approved):

OrlandoCo:
I added a change to the recieves stats. But I noticed that test are failing over a race condition.

PR #176, opened 2020-09-16
fix(sender): write on closed pipe err

tarrencev:
Handle `io.ErrClosedPipe` more gracefully.

We should simplify how we are closing senders. It should be triggered by the `p.pc.RemoveTrack` call, rather than the other way around.

### DIFF
--- a/pkg/sender.go
+++ b/pkg/sender.go
@@ -85,7 +85,10 @@ func (s *WebRTCSender) sendRTP() {
 			pkt = &newPkt
 
 			if err := s.track.WriteRTP(pkt); err != nil {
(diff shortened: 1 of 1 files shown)


OrlandoCo:
> We should simplify how we are closing senders. It should be triggered by the `p.pc.RemoveTrack` call, rather than the other way around.

According to this [comit](https://github.com/pion/webrtc/pull/1140) from Pion/Webrtc we should close the sender when a `io.EOF` it's recieved from the `webrtc.Track#Read()`, Currently now we are closing senders only on peer disconnecting /failing

OrlandoCo (approved):

PR #178, opened 2020-09-16
Added test for nairpair

obasajujoshua31:
added sending of rtcp nackpair from peer A to peerB through sfu

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/session_test.go
+++ b/pkg/session_test.go
@@ -2,16 +2,17 @@ package sfu
 
 import (
 	"context"
-	"math/rand"
-	"strconv"
-	"strings"
-	"testing"
-
 	"github.com/pion/ion-sfu/pkg/log"
+	"github.com/pion/rtcp"
 	"github.com/pion/sdp/v2"
 	"github.com/pion/transport/test"
 	"github.com/pion/webrtc/v3"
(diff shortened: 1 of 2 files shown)


pkg/session_test.go
```diff
@@ -571,3 +573,116 @@ func TestPeerBWithAudioAndVideoWhenPeerAHasAudioOnly(t *testing.T) {
...
+		SenderSSRC: trackB.SSRC(),
+		MediaSSRC:  trackB.SSRC(),
+		Nacks:      nil,
+	}
+
+	onTrackReceivedCtx, onTrackReceivedFunc := context.WithCancel(context.Background())
+
+	peerA.OnTrack(func(track *webrtc.Track, receiver *webrtc.RTPReceiver) {
+		ctx := context.Background()
+		videoReceiver := NewWebRTCVideoReceiver(ctx, WebRTCVideoReceiverConfig{
```
tarrencev, line 624: Instead of creating a Receiver here, we should call `sender.ReadRTCP() on track B's sender`

So on line 603, get the sender with

```
sender, err = remoteB.AddTrack(trackB)
```

and then here you can read the RTCP from it

tarrencev (changes requested):

PR #181, opened 2020-09-17
feat(receivers): refactor receivers

### LINKED ISSUE #177: We should simplify how we are closing senders. It should be triggered by the `p.pc.RemoveTrack` call, rather than the other way around.
> We should simplify how we are closing senders. It should be triggered by the `p.pc.RemoveTrack` call, rather than the other way around.

According to this [comit](https://github.com/pion/webrtc/pull/1140) from Pion/Webrtc we should close the sender when a `io.EOF` it's recieved from the `webrtc.Track#Read()`, Currently now we are closing senders only on peer disconnecting /failing

_Originally posted by @OrlandoCo in https://github.com/pion/ion-sfu/pull/176#issuecomment-693529630_

OrlandoCo:
#### Description
This PR it's for simplify the relation between Receivers->Routers->Senders

-Receivers
1. Will close on  remote track removed or Close called.
2. Simplify to only have one single struct for video/audio. 
3. Errors are handled in receivers and not bubbled up.

-Routers
1. Routers will only fwd data between senders and receivers.
1. Routers will close themselves on receiver closed.

-Senders (TODO)
1. Senders will close on peer removed or track removed
2. Senders feedbacks will be directly forwarded and not calculated in SFU (Discussion required) and can be opted out.

Tests (TODO)
1. Make some minor changes to reflect these changes.

Breaking Changes:
None, so far the api reminds the same.

#### Reference issue
Fixes #177

### DIFF
--- a/pkg/router.go
+++ b/pkg/router.go
@@ -63,21 +63,14 @@ func (r *Router) close() {
 }
 
 func (r *Router) start() {
-	rtpCh := r.receiver.ReadRTP()
 	defer r.close()
-	for {
-		select {
-		case pkt, opn := <-rtpCh:
-			if !opn {
-				return
-			}
-			// Push to sub send queues
-			r.mu.RLock()
-			for _, sub := range r.senders {
-				sub.WriteRTP(pkt)
-			}
-			r.mu.RUnlock()
+	for pkt := range r.receiver.ReadRTP() {
+		// Push to sub send queues
+		r.mu.RLock()
+		for _, sub := range r.senders {
+			sub.WriteRTP(pkt)
 		}
+		r.mu.RUnlock()
 	}
 }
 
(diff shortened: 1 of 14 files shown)


pkg/router.go
```diff
@@ -52,51 +50,34 @@ func (r *Router) AddSender(pid string, sub Sender) {
...
-			log.Errorf("r.receiver.ReadRTP err=%v", err)
-			continue
-		}
-
-		// Push to sub send queues
-		r.mu.RLock()
-		for _, sub := range r.senders {
-			sub.WriteRTP(pkt)
+		select {
+		case pkt, opn := <-rtpCh:
```
tarrencev, line 70: I think we can simplify here to
```
for pkt, opn := range rtpCh:
```

pkg/router.go
```diff
@@ -109,52 +90,47 @@ func (r *Router) subFeedbackLoop(pid string, sub Sender) {
...
-					sub.WriteRTP(bufferpkt)
-					continue
-				}
-
-				if routerConfig.MaxNackTime > 0 {
-					ln := atomic.LoadInt64(&r.lastNack)
-					if (time.Now().Unix() - ln) < routerConfig.MaxNackTime {
+	for {
+		select {
+		case pkt, opn := <-rtcpCh:
```
tarrencev, line 97: same as above
```
for pkt := range rtcpCh:
```

tarrencev, line 97: same as above
```
for pkt := range rtcpCh:
```

pkg/router_test.go
```diff
@@ -32,7 +32,7 @@ func TestRouter(t *testing.T) {
 	done := make(chan bool)
 	onTrackFired, onTrackFiredFunc := context.WithCancel(context.Background())
 	pubsfu.OnTrack(func(track *webrtc.Track, _ *webrtc.RTPReceiver) {
-		receiver := NewWebRTCVideoReceiver(ctx, WebRTCVideoReceiverConfig{}, track)
+		receiver := NewWebRTCReceiver(ctx, track).(*WebRTCReceiver)
```
tarrencev, line 35: Seems having the `Sender` `Receiver` interface doesn't really buy us much. Unless we plan to support other transports but could be awhile before that makes sense. Should we just use the `WebRTC*` types for now?

pkg/webrtctransport.go
```diff
@@ -281,21 +273,17 @@ func (p *WebRTCTransport) Close() error {
...
-			log.Errorf("Error reading RTCP %s", err)
-			continue
-		}
-
-		log.Tracef("sendRTCP %v", pkt)
-		err = p.pc.WriteRTCP([]rtcp.Packet{pkt})
-		if err != nil {
-			log.Errorf("Error writing RTCP %s", err)
+		select {
+		case pkt, opn := <-rtcpCh:
```
tarrencev, line 279:
```
for pkt := range rtcpCh:
```

tarrencev (approved):
minor comments otherwise lgtm! lets try to get the CI to pass before merging

pkg/router_test.go
OrlandoCo: I think they can help for mocking for uni testing, at the moment testing some parts like router becames complicated to test because we need to take care of peer signaling that almost become an e2e test.

pkg/router.go
OrlandoCo: Done in last commit.

PR #189, opened 2020-09-22
Add simulcast support

### LINKED ISSUE #57: simulcast support
integrate pion simulcast

- [ ] Update remb loop to add bw state to sender
- [ ] sscrc/rid/quality map exposed to sender
- [ ] Router forwards all simulcast packets to sender (may be no change)
- [ ] Sender inspects simulcast packet and forwards packets for stream best matching current sender bw
- [ ] Make sure client is sending abs-time extensions. (required for correct remb)
- [ ] Configure abs-time extension in offers

OrlandoCo:
#### Description

Look simulcast project.

#### Reference issue
Fixes #57


pkg/receiver.go
```diff
@@ -100,6 +111,35 @@ func (w *WebRTCReceiver) OnCloseHandler(fn func()) {
...
+}
+
+func (w *WebRTCReceiver) DeleteSender(pid string) {
+	w.Lock()
+	defer w.Unlock()
+	delete(w.senders, pid)
+}
+
+func (w *WebRTCReceiver) SetSimulcast(ssrc uint32) {
+	w.simulcast = true
```
tarrencev, line 127: this is set explicitly here but also derived from track.RID() above which is a bit confusing

pkg/simulcastsender.go
```diff
@@ -0,0 +1,214 @@
...
+			// Wait for a keyframe to sync new source
+			switch s.track.Codec().PayloadType {
+			case webrtc.DefaultPayloadTypeVP8:
+				vp8Packet := codecs.VP8Packet{}
+				if _, err := vp8Packet.Unmarshal(pkt.Payload); err == nil {
+					if vp8Packet.Payload[0]&0x01 == 0 {
+						// Packet is a keyframe
+						relay = true
+					}
+				}
```
tarrencev, line 93: Can we use the partition head checker?

```suggestion
				check := codecs.VP8PartitionHeadChecker{}
				if ok := check.IsPartitionHead(pkt); ok {
				        // Packet is a keyframe
					relay = true
				}
```

Same exists for other VP9/H264

pkg/simulcastsender.go
```diff
@@ -0,0 +1,214 @@
...
+		pkt.SSRC = s.simulcastSSRC
+		pkt.SequenceNumber = s.bSN
+		pkt.Timestamp = s.bTS
+		// Write packet to client
+		err := s.track.WriteRTP(pkt)
+		// Reset packet data
+		pkt.Timestamp = origTS
+		pkt.SSRC = origSSRC
+		pkt.SequenceNumber = origSeq
+		pkt.PayloadType = origPT
```
tarrencev, line 147: This is quite complex. I wonder if we can break it out into some helpers to make it easier to grok

tarrencev (commented):
some minor comments otherwise looks really good!

pkg/helpers.go
```diff
@@ -0,0 +1,127 @@
+package sfu
+
+import (
+	"encoding/binary"
+)
+
+// VP8Helper is a helper to get temporal data from VP8 packet header
```
tarrencev, line 7: Should we contribute this to https://github.com/pion/rtp/blob/master/codecs/vp8_packet.go ?

OrlandoCo: This just take care of the temporal data, it doesn't care of other flags or make a copy a of the vp8 payload for performance. So it's out of scope of the rtp/codecs. I can make a pull request to add tempral flags on vp8 codec, but whe should not use it on sfu for temporal info.

pkg/webrtctransport.go
```diff
@@ -53,51 +57,79 @@ func NewWebRTCTransport(ctx context.Context, session *Session, me MediaEngine, c
...
-
-		router := NewRouter(p.id, recv)
-		log.Debugf("Created router %s %d", p.id, recv.Track().SSRC())
-
-		p.session.AddRouter(router)
+		if router, ok := p.routers[track.ID()]; !ok {
+			if track.RID() != "" {
+				router = newRouter(p.id, cfg.router, SimulcastRouter)
+				go func() {
+					// Send 3 big remb msgs to fwd all the tracks
```
tarrencev, line 90: So we dont do any congestion control on the pub tracks right now?

tarrencev (approved):

pkg/sender.go
```diff
@@ -5,95 +5,94 @@ import (
...
+	onCloseHandler func()
 
 	once sync.Once
 }
 
 // NewWebRTCSender creates a new track sender instance
-func NewWebRTCSender(ctx context.Context, track *webrtc.Track, sender *webrtc.RTPSender) Sender {
+func NewWebRTCSender(ctx context.Context, id string, router Router, sender *webrtc.RTPSender) Sender {
 	ctx, cancel := context.WithCancel(ctx)
+	sender.Track()
```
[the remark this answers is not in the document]
OrlandoCo: no, a typo, will fixt it.

pkg/webrtctransport.go
OrlandoCo: yes, currently a remb loop is called per track, to calculate the bandwidth and later downgrade senders, but we can omit fwding the remb packet.

PR #196, opened 2020-09-28
feature/ion-sfu-mute

decanus:
#### Description

Allows single webrtc transports to be muted.

### DIFF
--- a/pkg/receiver.go
+++ b/pkg/receiver.go
@@ -52,6 +52,7 @@ type Receiver interface {
 type WebRTCReceiver struct {
 	ctx            context.Context
 	cancel         context.CancelFunc
+	mux            sync.RWMutex
 	buffer         *Buffer
 	track          *webrtc.Track
 	bandwidth      uint64
@@ -64,6 +65,8 @@ type WebRTCReceiver struct {
 	maxBandwidth uint64
 	feedback     string
 	wg           sync.WaitGroup
+
+	muted bool
 }
 
 // WebRTCVideoReceiverConfig .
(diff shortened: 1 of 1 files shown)


pkg/receiver.go
```diff
@@ -451,6 +478,11 @@ func startAudioReceiver(w *WebRTCReceiver, wStart chan struct{}) {
 				continue
 			}
 
+			// @TODO SHOULD THIS BE HERE?
+			if w.muted {
+				continue
+			}
+
```
OrlandoCo, line 485: Since a webrtc stream must have a continuos sequence number and timestamps, ommiting packets like this, will cause that the client fwd many nacks or freeze once is unmuted.

The mute must be done in client side, or in server side but the server must keep sending silent or black frames packets and resync once is unmuted.

OrlandoCo (changes requested):

pkg/receiver.go
decanus: @OrlandoCo yeah this was very rough. I assumed we will need to send empty packets. Another thing I was thinking about doing is maybe renegotiating and not adding a receiving audio track for the given peer. And allow that to be changed. What do you think?

tarrencev: @OrlandoCo is that still the case with DTX? I think we can server side mute if DTX is supported. I'm not sure if that requires any changes on the sfu side

OrlandoCo: @tarrencev As far as I know DTX is only for audio and I'm not sure if it solve this problem, need to investigate.

@decanus You can mute or remove tracks on client side without problems, renogotiation will be done on remove tracks, and on mute the client will automatically forward muted packets.

decanus: So the problem is for my use case, there is a room member that is the admin and can decide to mute other members. In which case their track would be removed. What do you think is the best method of doing this @OrlandoCo?

OrlandoCo: @decanus You can use your own signaling or datachannels to force the user to mute his track.

decanus: @OrlandoCo I guess that works. But how could I ensure its enforced server side?

OrlandoCo: @decanus At the moment is not supported, please open an issue so we can discuss the best way how to solve this.

decanus: https://github.com/pion/ion-sfu/issues/198

PR #207, opened 2020-10-01
feat(sfu): Add data-channel sub api

OrlandoCo:
#### Description

This PR draft it's to discuss the API to control the SFU. This API is intended only to control receiving tracks, like muting a track or change the layer of a track.

This API is not designed  to control or remove tracks from another users.


pkg/api.go
```diff
@@ -0,0 +1,85 @@
...
+	// Server: Send confirmation that stream is un-muted
+	unmuteCommand = "unmute"
+
+	// Best Quality Command
+	// Direction: Client -> SFU -> Client
+	// Client: Send intent to get best quality of the stream
+	// Server: Send confirmation that stream is in best quality,
+	// send an error if stream does not support simulcast/SVC
+	// NOTE: client may not get best quality is bw is a limitation.
+	forceBestQuality = "bestQuality"
```
tarrencev, line 36: i'm not sure if we want to allow the client to "force". i feel like they should request but can be overridden by the sfu's congestion control. for example, if a speaker is active and large screen, i would like their high res but wouldn't want to force it if it degrades the stream. i feel like the client should always be a soft preference. but maybe i am missing a use case

tarrencev: oh i just saw your comment `NOTE: client may not get best quality is bw is a limitation`. maybe it could be `preferBestQuality`

pkg/api.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/pion/ion-sfu/pkg/log"
+	"github.com/pion/webrtc/v3"
+)
+
+// All available data channels commands for SFU, may reflect only changes on caller,
+// this is not intended to force remove,mute, etc. to other publisher tracks.
+// Data channel commands are received in ion-sfu data channel. So it must be considered
+// a reserved name
+const (
```
tarrencev, line 14: generally i would prefer an api that is constraints oriented, similar to `getUserMedia`. this is because the client has a fixed number of pixels to fill, so it is easier to reason about the track they want by saying "the thumbnail view is 320x240, send me the track with the closest resolution`. otherwise the api is kind of opaque, i dont really know what i am asking for. the resolution logic might be a little complex though

OrlandoCo: Getting the tracks width and height would require inpecting the packets, signaling them is not warranty of the recieved tracks use them this apply more in simulcast/SVC where the substreams are hardcoded.

So if we add a command in client SDK like ```getRemoteMedia(video:"high","medium","low" ,audio:boolean)```, we could just use high, medium or low instead resolutions, and internally translate them to more simple commands for the SFU.

pkg/webrtctransport.go
```diff
@@ -77,9 +76,18 @@ func NewWebRTCTransport(ctx context.Context, session *Session, me webrtc.MediaEn
 	session.AddTransport(p)
 
 	pc.OnTrack(func(track *webrtc.Track, receiver *webrtc.RTPReceiver) {
-		log.Debugf("Peer %s got remote track id: %s ssrc: %d rid :%s label: %s", p.id, track.ID(), track.SSRC(), track.RID(), track.Label())
+		log.Debugf("Peer %s got remote track id: %s ssrc: %d rid :%s streamID: %s", p.id, track.ID(), track.SSRC(), track.RID(), track.Label())
 		recv := NewWebRTCReceiver(ctx, track, cfg.router)
 
+		// Set transport label
+		if p.streamID == "" {
```
tarrencev, line 83: Does this logic assume that a transport can only publish a single "MediaStream"?

pkg/api.go
tarrencev: i really like that api actually. we could have a single command that accept a payload like:
```
{
     id: "....",
     video:"high"|"medium"|"low"|"none" ,
     audio:boolean
}
```

pkg/webrtctransport.go
OrlandoCo: Yes, it  can be moved to routers, but I could not find a use-case to publish multiple media streams from a single peer.

tarrencev: one use case is screen sharing + camera

tarrencev (approved):

PR #214, opened 2020-10-06
add allrpc

adwpc:
#### Description
* allrpc support `grpc` or `jsonrpc` or both
* you can do  a 1:N load-tesing by ion-load-tool(grpc)
* you can use pub-sub-in-browsor(jsonrpc) to join the room and see the video


tarrencev:
i feel like it might be possible to modularize jsonrpc/grpc so it isn't necessary to duplicate all of the logic

adwpc:
> i feel like it might be possible to modularize jsonrpc/grpc so it isn't necessary to duplicate all of the logic

Good idea

PR #219, opened 2020-10-09
Modularize peer signaling

billylindeman:
#### Description
- Refactors peer signaling to be abstracted from the actual transport / serialization 
- provides a jsonPeer wrapper that acts as a jsonrpc2 handler 

This should make it easier to add new transports and additional layers while keeping the sfu api consistent.

TODO:
- [x] Create grpcPeer interface


adwpc:
Similar to mine allrpc，but will support grpc/jsonrpc both

pkg/peer.go
```diff
@@ -0,0 +1,154 @@
...
+	OfferChan chan *webrtc.SessionDescription
+}
+
+// NewPeer creates a new Peer for signaling with the given SFU
+func NewPeer(sfu *SFU) Peer {
+	return Peer{
+		sfu:       sfu,
+		Done:      make(chan bool),
+		IceChan:   make(chan *webrtc.ICECandidate),
+		OfferChan: make(chan *webrtc.SessionDescription),
```
tarrencev, line 34: I would prefer to use plain callbacks here. Since we don't need to worry about concurrency so chans add a bunch of overhead. Otherwise looking good!

PR #220, opened 2020-10-10
feat(signaling): join answer + offer

tarrencev:
Improvement to single pc signaling based on suggestion by @jech

When doing the initial negotiate of the peer connection (on join), the sfu responds with an answer to the remote peer's offer as well as a new offer including any media for existing peers in the session. We then start sending RTP when the sfu receives an answer from the remote peer (this part still need to be implemented).


tarrencev:
We are going to take a different approach. Senders will start in the muted state and be unmuted by the client when they want the media.

PR #223, opened 2020-10-11
feat(signal): modularize server

tarrencev:


billylindeman (approved):
Nice, I was thinking about doing something similar this week 👍 

This will make ion-sfu-cluster a bit easier to build. I'm going to spend some time this week working on it

PR #232, opened 2020-10-18
fix: Remove the duplication of addEventListener

### LINKED ISSUE #231: [fix] Remove duplication of addEventListener
## Summary

In echotest example's `index.html`, The lines from 172 to 191 seem to be exactly the same with the lines from 206 to 227. 
As they are in the same scope, I guess the same handler is being assigned twice. 

## Motivation

Tried testing after removing the lines from 172 to 191, and it works fine.

## Describe alternatives you've considered

Remove the lines from 172 to 191. 

## Additional context

If this duplication was intended, sorry for bothering.

JonghunBok:
#### Description
Remove the lines from 172 to 191.

#### Reference issue
Fixes #231

### DIFF
--- a/examples/echotest/index.html
+++ b/examples/echotest/index.html
@@ -169,27 +169,6 @@
         const pc = new RTCPeerConnection(config)
         const epc = new RTCPeerConnection(config)
 
-        socket.addEventListener('message', async (event) => {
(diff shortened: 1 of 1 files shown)


tarrencev:
thanks for the fix!

PR #233, opened 2020-10-18
feat(add allrpc)

adwpc:
#### Description
* add allrpc, can use jsonrpc/grpc/both now, easy to do load-testing with ion-load-tool  and echotest at the same time.
* enable sending packet


cgojin (commented):
great! I think so, pub using json-rpc in browser, and ion-avp sub using grpc.

billylindeman (approved):
LGTM! Much cleaner :D

PR #234, opened 2020-10-18
fix log

### LINKED ISSUE #167: remove logging `fix` from config
I dont think we need the `fix` parameter here to be configurable:

https://github.com/pion/ion-sfu/blob/b166816adeaef49cdf4cd719458b57163146d46a/config.toml#L36

Since it is known statically, so it should just be hard coded. Unless I am missing something? When would someone want to configure this?

adwpc:
#### Description
* change log to ion-log
* hardcode log filter
* rm unused log config

#### Reference issue
Fixes #167


cgojin (commented):
great!

adwpc:
Error: /bin/tar: ../../../go/pkg/mod/golang.org/x/net@v0.0.0-20201016165138-7b1cca2348c0/html/charset/charset_test.go: Cannot open: File exists
not know how to fix these golint-cli errors
merge first

PR #245, opened 2020-10-21
feat: add combined grpc/json-rpc service for disk/browser examples

jvanveen:
#### Description
It would be great if browser examples can be combined with examples that are run from disk.
The pub-from-disk example uses grpc, while the browser uses json-rpc. This PR tries to combine 
both signalling methods, so I can play an example movie from disk to the browser. Accompanying 
PR in examples: https://github.com/pion/ion-examples/pull/34

I'm rather new to Golang and Ion-Sfu. Please let me know if this PR needs more work. :)
Edit: Just read the contributing guidelines. Will try to polish things up later, so all tests pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ vendor
 *.pem
 bin
 *.generated.go
+ion-sfu
--- a/README.md
+++ b/README.md
@@ -1,61 +1,41 @@
-<h1 align="center">
-  <br>
-  Ion SFU
-  <br>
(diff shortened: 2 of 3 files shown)


tarrencev:
hi @jvanveen thank you for your contribution. i believe this was just merged recently, there is now an `allrpc` signaling server

jvanveen:
Hi @tarrencev, that's great! I'll have a look

PR #253, opened 2020-10-24
Remove the config assignment not being used

### LINKED ISSUE #252: [fix] Remove unused config variable
## Summary

The variable `config` was assigned, but being used. 
In this example, as signaling process is being done via ion-sfu server, the `config` seems to serve no purpose.

## Motivation

As `config` doesn't play a role in this use case, it can confuse the readers of this example.
Removing it can clarify the role of `ion-sfu`.

## Describe alternatives you've considered

Remove the `config` assignment part.

## Additional context

By the way, thank you for writing such an on-point and readable example. 
This one helped me a lot.

JonghunBok:
#### Description

The variable `config` was assigned, but being used. 
In this example, as signaling process is being done via ion-sfu server, the `config` seems to serve no purpose.

If this was intended, sorry for bothering.

#### Reference issue
Fixes #252

### DIFF
--- a/examples/echotest/index.html
+++ b/examples/echotest/index.html
@@ -300,14 +300,6 @@
       /* eslint-env browser */
       const joinBtns = document.getElementById("start-btns");
 
-      const config = {
-        iceServers: [
-          {
-            urls: "stun:stun.l.google.com:19302",
-          },
-        ],
-      };
-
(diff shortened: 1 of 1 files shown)


tarrencev:
awesome thanks!

PR #255, opened 2020-10-24
fix(sender): start senders after answer is received

tarrencev:
a race exists where rtp can arrive at the remote client before they have handled the sdp. this delays sending the rtp until the sfu has received an answer from the remote client

### DIFF
--- a/pkg/webrtctransport.go
+++ b/pkg/webrtctransport.go
@@ -2,12 +2,14 @@ package sfu
 
 import (
 	"context"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/lucsky/cuid"
 	log "github.com/pion/ion-log"
 	"github.com/pion/rtcp"
+	"github.com/pion/sdp/v2"
 	"github.com/pion/webrtc/v3"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/webrtctransport.go
```diff
@@ -2,12 +2,14 @@ package sfu
...
 import (
 	"context"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/lucsky/cuid"
 	log "github.com/pion/ion-log"
 	"github.com/pion/rtcp"
+	"github.com/pion/sdp/v2"
```
OrlandoCo, line 12: Can we switch to v3?

OrlandoCo:
Lgtm!

OrlandoCo (approved):

pkg/webrtctransport.go
```diff
@@ -186,7 +190,37 @@ func NewWebRTCTransport(ctx context.Context, session *Session, me MediaEngine, c
 
 // CreateOffer generates the localDescription
 func (p *WebRTCTransport) CreateOffer() (webrtc.SessionDescription, error) {
-	return p.pc.CreateOffer(nil)
+	offer, err := p.pc.CreateOffer(nil)
```
OrlandoCo, line 193: Create offer already return a parsed offer, but it's private, I wonder if we should make it public

PR #261, opened 2020-10-27
Fix nack locking buffer writes

OrlandoCo:
#### Description

This PR solves the issue when many nacks blocks the buffer

### DIFF
--- a/pkg/buffer.go
+++ b/pkg/buffer.go
@@ -2,6 +2,7 @@ package sfu
 
 import (
 	"sync"
+	"sync/atomic"
 	"time"
 
 	log "github.com/pion/ion-log"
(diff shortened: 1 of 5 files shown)


pkg/buffer.go
```diff
@@ -198,18 +200,16 @@ func (b *Buffer) buildReceptionReport() rtcp.ReceptionReport {
...
 	return rr
 }
 
 func (b *Buffer) setSenderReportData(rtpTime uint32, ntpTime uint64) {
-	b.mu.Lock()
-	defer b.mu.Unlock()
-	b.lastSRRTPTime = rtpTime
-	b.lastSRNTPTime = ntpTime
-	b.lastSRRecv = time.Now().UnixNano()
+	atomic.StoreUint32(&b.lastSRRTPTime, rtpTime)
```
tarrencev, line 210: should we store this in a struct and use `StoreValue`? so they are all atomically set together

PR #269, opened 2020-10-29
Remove context

OrlandoCo:
Remove context

### DIFF
--- a/pkg/buffer.go
+++ b/pkg/buffer.go
@@ -1,7 +1,6 @@
 package sfu
 
 import (
-	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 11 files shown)


billylindeman:
I'm confused why you're removing context, if anything I feel like we should add more and actually use them properly (for example a context on every session / every PC in the session?)

OrlandoCo:
@billylindeman  Context should be request scoped and not being keeped as reference, also cancel make sense where for some reason the caller should cancel the downstream flow, this not make sense in this scenario, where peers in SFU depends on remotes. In case a session needs to be shutdown this need to be in a more controlled scenerio and not cancelling a parent context that would lead to many bugs.

Also the context tree get bigger for every peer connected and causing many contention on hot paths like read/write packets. In conclusion using context the way it was used in sfu was prone to bugs and not being really useful and redundant.

PR #271, opened 2020-10-30
feat(test): e2e kite testing

tarrencev:


adwpc:
seems too old, just close

PR #272, opened 2020-10-30
fix: lock optizations + debugging

tarrencev:


pkg/receiver.go
```diff
@@ -234,10 +237,12 @@ func (w *WebRTCReceiver) readSimulcastRTCP(rid string) {
 func (w *WebRTCReceiver) writeRTP() {
 	for pkt := range w.rtpCh {
 		w.RLock()
-		for _, sub := range w.senders {
+		senders := w.senders
+		w.RUnlock()
```
OrlandoCo, line 241: This will lead to a race, since only get a reference to the map

pkg/session.go
```diff
@@ -46,9 +46,7 @@ func (r *Session) RemoveTransport(tid string) {
 
 // AddRouter adds a router to transports
 func (r *Session) AddRouter(router Router, rr *receiverRouter) {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-	for tid, t := range r.transports {
+	for tid, t := range r.Transports() {
```
OrlandoCo, line 49: Also can lead to a race if lock is removed

OrlandoCo (changes requested):

pkg/receiver.go
```diff
@@ -234,10 +237,10 @@ func (w *WebRTCReceiver) readSimulcastRTCP(rid string) {
 func (w *WebRTCReceiver) writeRTP() {
 	for pkt := range w.rtpCh {
 		w.RLock()
+		defer w.RUnlock()
```
OrlandoCo: Must not be defered since it's inside a loop

OrlandoCo (approved):

PR #302, opened 2020-11-09
fix(sfu): Add locks in Peer

OrlandoCo:
#### Description

Add locks in Peer, to avoid racings

### DIFF
--- a/pkg/webrtctransport.go
+++ b/pkg/webrtctransport.go
@@ -146,8 +146,8 @@ func (p *WebRTCTransport) SetRemoteDescription(desc webrtc.SessionDescription) e
 	switch desc.Type {
 	case webrtc.SDPTypeAnswer:
 		p.router.SetExtMap(pd)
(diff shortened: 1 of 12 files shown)


tarrencev (approved):

pkg/webrtctransport.go
```diff
@@ -222,10 +219,7 @@ func (p *WebRTCTransport) OnICECandidate(f func(c *webrtc.ICECandidate)) {
 
 // OnNegotiationNeeded handler
 func (p *WebRTCTransport) OnNegotiationNeeded(f func()) {
-	debounced := debounce.New(100 * time.Millisecond)
```
tarrencev, line 225: if we remove this, will we get a negotiate for each track?

OrlandoCo: No, because now we wait unitl all tracks are added to the peer before negotiation is fired.

PR #328, opened 2020-12-08
feat(SFU): add sender reports

OrlandoCo:
#### Description

Add reports to downtracks

### DIFF
--- a/pkg/sfu/downtrack.go
+++ b/pkg/sfu/downtrack.go
@@ -80,11 +80,11 @@ func NewDownTrack(c webrtc.RTPCodecCapability, r Receiver, peerID, id, streamID 
 func (d *DownTrack) Bind(t webrtc.TrackLocalContext) (webrtc.RTPCodecParameters, error) {
 	parameters := webrtc.RTPCodecParameters{RTPCodecCapability: d.codec}
 	if codec, err := codecParametersFuzzySearch(parameters, t.CodecParameters()); err == nil {
-		d.bound.set(true)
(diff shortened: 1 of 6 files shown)


robin-raymond:
This will help ensure SSRCs are synchronized. The other piece of information that's vital is ensuring the SDES/CNAME entry for audio and video is delivered to ensure the coordination of the SSRCs for synchronization purposes. Is that already completed elsewhere? It would not necessarily be present in the SRs RTCP packets.

OrlandoCo:
> This will help ensure SSRCs are synchronized. The other piece of information that's vital is ensuring the SDES/CNAME entry for audio and video is delivered to ensure the coordination of the SSRCs for synchronization purposes. Is that already completed elsewhere? It would not necessarily be present in the SRs RTCP packets.

No yet, I will implement it now in this PR !

pkg/sfu/downtrack.go
```diff
@@ -166,6 +177,13 @@ func (d *DownTrack) OnCloseHandler(fn func()) {
 	d.onCloseHandler = fn
 }
 
+func (d *DownTrack) OnBind(f func()) {
+	debounced := debounce.New(250 * time.Millisecond)
```
tarrencev, line 181: why do we need to debounce here?

tarrencev (approved):

pkg/sfu/downtrack.go
OrlandoCo: Im fixing that now, to not use debouce, and just send reports for stream id

PR #343, opened 2020-12-17
Feat/buckets

OrlandoCo:
#### Description

Reduce slice allocation in rtp/rtcp readers.


pkg/buffer/bucket.go
```diff
@@ -0,0 +1,133 @@
...
+			b.nacker.remove(sn)
+		}
+		return b.set(sn, pkt)
+	}
+	diff := sn - b.headSN
+	b.headSN = sn
+	for i := uint16(1); i < diff; i++ {
+		b.step++
+		if b.nacker != nil {
+			b.counter++
```
tarrencev, line 46: nit: might make more sense for counter to be a member on `nacker`

pkg/buffer/factory.go
```diff
@@ -0,0 +1,72 @@
...
+	audioPool   *sync.Pool
+	rtpBuffers  map[uint32]*Buffer
+	rtcpReaders map[uint32]*RTCPReader
+}
+
+func NewBufferFactory() *Factory {
+	return &Factory{
+		videoPool: &sync.Pool{
+			New: func() interface{} {
+				return NewBucket(2*1000*1000, true)
```
tarrencev, line 22: could you add a comment about the intuition around this constant and the others in the pr?

tarrencev (approved):
awesome stuff! a lot of code but it looks good. i think best to test it out

adwpc:
Awsome!

OrlandoCo:
@tarrencev  I fixed your request and added h264 keyframe detection, credits to @jech I think is safe to merge now

jech:
In Galène, the keyframe detection code returns a ternary result — yes, no, couldn't be determined.  In Ion-SFU, the code returns a binary result.

So I think that the behaviour at line 205 should be changed somewhat: this is a case that is probably illegal (non-simple NALU within a STAP), and the Galène code simply gives up in this case, returning the « couldn't be determined » result.  In Ion-SFU, you should probably continue walking the STAP, in case you find a keyframe later.

I suggest simply removing the line that says

    return false

and perhaps replace it with some logging (so we know whether that case actually happens in practice).

pkg/sfu/helpers.go
```diff
@@ -157,6 +157,73 @@ func setVP8TemporalLayer(pl []byte, s *DownTrack) (payload []byte, skip bool) {
...
+			}
+			if offset >= int(length) {
+				return false
+			}
+			n := payload[i+offset] & 0x1F
+			if n == 5 {
+				return true
+			} else if n >= 24 {
+				// is this legal?
+				return false
```
jech, line 205: This line should probably be removed — in Galène, it gives up at this point, but Ion-SFU is different, it should continue walking the STAP.

OrlandoCo: done, thank you!

Issue #347, opened 2020-12-28
Some server side ICE candidates are lost


tmiv:
### Your environment.
  - Version: 3ae96731c273ab2d459e9c42793d928feec0a84f
  - Browser: This is an issue between the signal server and the SFU. Can be reproduced without a browser.
  - Other Information - I have an incoming pull request to fix this. Essentially the issue is due to setting the remote description before setting the ice callbacks.

### What did you do?
This is most observable when nat1to1 and icelite are enabled in the config.toml and no webrtc.iceservers are passed into the config. When that is the case you'd expect only a single ice candidate to get sent by the sfu. When this config is used the ice candidate is missed 1 out of 10 times.  

If this debug test code is added
```
--- a/pkg/sfu/peer.go
+++ b/pkg/sfu/peer.go
@@ -80,6 +80,10 @@ func (p *Peer) Join(sid string, sdp webrtc.SessionDescription) (*webrtc.SessionD
                return nil, fmt.Errorf("error creating transport: %v", err)
        }
 
+       p.publisher.OnICECandidate(func(c *webrtc.ICECandidate) {
+               log.Debugf("Ice Func Not yet Set!")
+       })
+
        p.session.AddPeer(p)
 
        log.Infof("peer %s join session %s", p.id, sid)
@@ -131,12 +135,16 @@ func (p *Peer) Join(sid string, sdp webrtc.SessionDescription) (*webrtc.SessionD
        p.publisher.OnICECandidate(func(c *webrtc.ICECandidate) {
                log.Debugf("on ice candidate called")
                if c == nil {
+                       log.Debugf("ice candidate nil")
                        return
                }
 
                if p.OnIceCandidate != nil {
                        json := c.ToJSON()
+                       log.Debugf("ice candidate %s", json)
                        p.OnIceCandidate(&json, publisher)
+               } else {
+                       log.Debugf("ice candidate func nul")
                }
```
the statement 'Ice Func Not yet Set!' will fire in the condition when the ice candidate is lost.

```
[2020-12-28 18:24:36.521] [INFO] [89][peer.go][Join] => peer ckj8w5ew00001cv85r80wgvjy join session -MEBmPDPj06WJ683FrnG
[2020-12-28 18:24:36.524] [DEBUG] [72][publisher.go][func3] => ice connection state: checking
[2020-12-28 18:24:36.525] [INFO] [96][peer.go][Join] => peer ckj8w5ew00001cv85r80wgvjy send answer
[...]

PR #354, opened 2021-01-01
Reorder calls to fix ice candidate not set when ice starts to trickle

tmiv:
#### Description
The join before setting of the OnIceCandidate callback was causing an Ice candidate to not get sent in a very slim (1 out of 10) non deterministic manner.

#### Reference issue
I did not create an issue for this, but I can if you'd prefer. Its very similar to #347


OrlandoCo:
@tarrencev this should apply for all the other grpc signalers.

PR #356, opened 2021-01-02
feat: metrics port

tarrencev:
port of #341 to new buffer


pkg/stats/interceptor.go
```diff
@@ -0,0 +1,272 @@
...
+				}
+
+				calculateLatestMinMaxSenderNtpTime := func(cname string) (minPacketNtpTimeInMillisSinceSenderEpoch uint64, maxPacketNtpTimeInMillisSinceSenderEpoch uint64) {
+					if len(cname) < 1 {
+						return
+					}
+					i.RLock()
+					defer i.RUnlock()
+
+					for _, s := range i.streams {
```
tarrencev, line 172: should we keep a mapping of cname to streams to avoid these iterations?

PR #357, opened 2021-01-03
patch: Allowing Peer Id to be set externally

dinumathai:
#### Description
Allowing Peer Id to be set externally. This will allow logs to be more traceable.

### DIFF
--- a/pkg/sfu/peer.go
+++ b/pkg/sfu/peer.go
@@ -57,12 +57,17 @@ func NewPeer(provider SessionProvider) *Peer {
 
 // Join initializes this peer for a given sessionID (takes an SDPOffer)
 func (p *Peer) Join(sid string, sdp webrtc.SessionDescription) (*webrtc.SessionDescription, error) {
+	pid := cuid.New()
+	return p.JoinWithID(sid, sdp, pid)
+}
+
+// JoinWithID initializes this peer for a given sessionID (takes an SDPOffer). Peer Id can be set externally for debugging
+func (p *Peer) JoinWithID(sid string, sdp webrtc.SessionDescription, pid string) (*webrtc.SessionDescription, error) {
 	if p.publisher != nil {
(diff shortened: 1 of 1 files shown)


tarrencev:
hi @dinumathai thank you for your pr. for now, i would prefer not to merge this. if it's primary use case is for debugging then it is better to create a debug build and use that

dinumathai:
Externalizing the peerId will also give more control to signaling server.
Use case : Admin of a call want to kick out a user. This can be done only if the Admin knows the peer id.

tarrencev:
@dinumathai can you instead call `Close` on the reference to `sfu.Peer` for that use case? we can also expose the ID if that helps. @OrlandoCo any ideas?

OrlandoCo:
@dinumathai @tarrencev  well not sure of the benefits from doing this, since if you set the peerID then you will need a way to relate the peerID with the streams IDs, but once you do that then why not keep with the stream ID??, is the taking that everyone does for webrtc just relate the streams ID to an external user ID, then you can set an endpoint to take a session ID and stream ID and you can decide between end the stream or fully remove the peer.

dinumathai:
Thanks, That solves the problem "Admin of a conference want to kick out a user".

Could you please suggest something for the log issue. I was thinking adding a "generateid-sid-ExternalIserID"
as peer id, to make tracing easy.
> Making logs to be more traceable. Bottom line - I must be able to run a grep on the log and get all the logs of signaling server and the sfu.

dinumathai:
Please think about this one more time. With an internal id in log we need to guess a lot.

billylindeman:
I think we should revisit this soon. It's been on my todo list to add support for this as well.  SessionID's and StreamID's are correlated between call logs and peer-id's are a sore thumb at the moment.

dinumathai:
Thank you @adwpc for https://github.com/pion/ion-sfu/pull/423

PR #381, opened 2021-01-19
fix-pubsub-allrpc

### LINKED ISSUE #344: Examples use track.onremovetrack but should use stream.onremovetrack
### Your environment.
  - Version: macOS 10.15.7
  - Browser: Firefox 84.0
  - Other Information - fixed by changing to stream.onremovetrack and checking event for track kind before removing the video element (because stream.onremovetrack is called twice)

### What did you do?
Tested with pubusbtest, open two browser tabs, start publishing on both, close a browser tab and the remotes video element belonging to the closed browser tab is frozen and not removed.

### What did you expect?

The video element should be removed correctly when the other party leaves the page or closes the browser tab.

### What happened?

The video element remains on the page and shows a frozen picture.

### LINKED ISSUE #345: Pubsubtest example creates multiple video elements for the same stream in chrome
### Your environment.
  - Version: 10.15.7
  - Browser: Chrome 87.0.4280.88
  - Other Information

### What did you do?
Tested with pubsubtest example, used with multiple browsers to create several video streams. Wait a little, but happened as soon as 30 seconds one time.

### What did you expect?
Every stream should be associated with only a single video element.

### What happened?
The onumute event is fired multiple times in Chrome which causes creation of additional video elements.

adwpc:
#### Description
fix(pubsub html add multi player)
feat(allrpc add metrics)
fix(gallerytest html add muti player)
fix(Concurrency-grpc-stream-Send in different goroutines)
fix(open too many files error by setting ice.MulticastDNSModeDisabled)

#### Reference issue
Fixes #345 
Fixes #344

### DIFF
--- a/cmd/signal/grpc/server/server.go
+++ b/cmd/signal/grpc/server/server.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"sync"
 
 	log "github.com/pion/ion-log"
 	pb "github.com/pion/ion-sfu/cmd/signal/grpc/proto"
@@ -42,6 +43,7 @@ func NewServer(sfu *sfu.SFU) *SFUServer {
 // If the client closes this stream, the webrtc stream will be closed.
 func (s *SFUServer) Signal(stream pb.SFU_SignalServer) error {
 	peer := sfu.NewPeer(s.SFU)
+	var streamLock sync.Mutex
 	for {
 		in, err := stream.Recv()
 
@@ -68,11 +70,13 @@ func (s *SFUServer) Signal(stream pb.SFU_SignalServer) error {
 			var offer webrtc.SessionDescription
 			err := json.Unmarshal(payload.Join.Description, &offer)
 			if err != nil {
+				streamLock.Lock()
 				err = stream.Send(&pb.SignalReply{
 					Payload: &pb.SignalReply_Error{
 						Error: fmt.Errorf("join sdp unmarshal error: %w", err).Error(),
 					},
 				})
+				streamLock.Unlock()
 				if err != nil {
 					log.Errorf("grpc send error %v ", err)
 					return status.Errorf(codes.Internal, err.Error())
@@ -85,6 +89,7 @@ func (s *SFUServer) Signal(stream pb.SFU_SignalServer) error {
 				if err != nil {
 					log.Errorf("OnIceCandidate error %s", err)
 				}
+				streamLock.Lock()
 				err = stream.Send(&pb.SignalReply{
 					Payload: &pb.SignalReply_Trickle{
 						Trickle: &pb.Trickle{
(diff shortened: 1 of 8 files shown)


cmd/signal/grpc/server/server.go
```diff
@@ -42,6 +43,7 @@ func NewServer(sfu *sfu.SFU) *SFUServer {
 // If the client closes this stream, the webrtc stream will be closed.
 func (s *SFUServer) Signal(stream pb.SFU_SignalServer) error {
 	peer := sfu.NewPeer(s.SFU)
+	var streamLock sync.Mutex
```
tarrencev, line 46: why is this necessary?

pkg/sfu/sfu.go
```diff
@@ -123,6 +124,7 @@ func NewWebRTCTransportConfig(c Config) WebRTCTransportConfig {
 	if len(c.WebRTC.Candidates.NAT1To1IPs) > 0 {
 		w.setting.SetNAT1To1IPs(c.WebRTC.Candidates.NAT1To1IPs, webrtc.ICECandidateTypeHost)
 	}
+	w.setting.SetICEMulticastDNSMode(ice.MulticastDNSModeDisabled)
```
tarrencev, line 127: should we make this configurable?

cmd/signal/grpc/server/server.go
adwpc: This is the error when i comment this lock:
`[2021-01-22 17:57:42.545] [ERROR] [132][peer.go][func4] => oniceconnectionstatechange error rpc error: code = Internal des
c = transport: transport: the stream is done or WriteHeader was already called`
The reason is:
https://github.com/grpc/grpc-go/blob/master/Documentation/concurrency.md#streams

pkg/sfu/sfu.go
adwpc: Not required in most scenarios
Disable mdns can save many file descriptors

adwpc: Would like to make this configurable if someone need :)

cmd/signal/grpc/server/server.go
tarrencev: got it. i have seen that error. can we embed the lock in the sever instead?

```
type SFUServer struct {
	pb.UnimplementedSFUServer
        sync.Mutex
	SFU *sfu.SFU
}
```

adwpc: Good idea. 
there is only one recv goroutine, so we only need a send lock.

pkg/sfu/sfu.go
adwpc: made it configuabe

tarrencev (approved):

pkg/sfu/sfu.go
OrlandoCo: @tarrencev @adwpc  There can be some security implication disabling the dns multicast for ice, so lets keep it enabled as default.

adwpc: OK

PR #382, opened 2021-01-19
feature/data-channel-middleware

### LINKED ISSUE #365: Data channel Middleware
Right now we support data channel fanout. It would be great to support a middleware interface for processing data channel content. This will allow for more complex data channel usage (i.e. processing inputs before forwarding to subs)

decanus:
#### Description

This is the first experiment with @tarrencev to look at data channel middleware.

#### Reference issue
Fixes #365


decanus:
The current interface for the function is as follows:

```golang
func(origin string, msg webrtc.DataChannelMessage, outputs map[string]*webrtc.DataChannel)
```

I think that I would prefer this:

```golang
func(origin, label string, msg webrtc.DataChannelMessage)
```

And allow the user to gather the data-channels sperately, in some cases the user may not need any outputs and we would be doing useless work.

tarrencev:
I would like an interface similar to this

```
dc := sfu.NewDatachannel()
// A good base middleware stack
dc.Use(middleware.Auth)
dc.OnMessage("notification", func(ctx, datachannel in, []datachannel out) {
  // do stuff
})
```
This would allow us to accomplish the ping/pong use case as well as msg fan out to subs. I think if we have users gather datachannels out of band it could result in buggy behavior. We can update the data structure to fit our access pattern. If we fully control the function context we can have more control of the channel life cycle, i.e. handle open/close state for the user

tarrencev:
Added in #388. Docs here: https://github.com/pion/ion-sfu/tree/master/pkg/middlewares/datachannel

PR #388, opened 2021-01-22
Add data channel middlewares

OrlandoCo:
#### Description

Add datachannels middlewares

WIP


OrlandoCo:
@decanus  @tarrencev  Please take a look of this implementation of the datachannel middlewares this will allow you to set a middleware chain on datachannels,  we can give access to session and it can be helpul to set an admin middleware to remove/silence users paired with an auth middleware, the API is straightforward to use and similar to a http middleware:

The way to add  a middleware chain is as follows:
```go 
	s := sfu.NewSFU(conf)
	dc := s.NewDatachannel(sfu.APIChannelLabel)
	dc.Use(datachannel.KeepAlive(5*time.Second), datachannel.SubscriberAPI)
        // Here its possible to fanout the message or act on processed msg or context values
        // If it's wanted we can give access to session to control other peers
        dc.OnMessage(func(ctx context.Context, msg webrtc.DataChannelMessage, in *webrtc.DataChannel, out []*webrtc.DataChannel) {
	})
```
This is the basic way to define a middleware:
```go
func SubscriberAPI(next sfu.MessageProcessor) sfu.MessageProcessor {
	return sfu.ProcessFunc(func(ctx context.Context, args sfu.ProcessArgs) {
              next.Process(ctx,args)
        }
}
```

tarrencev (approved):

PR #390, opened 2021-01-23
feat(turn): lt creds support

tarrencev:


pkg/sfu/turn.go
```diff
@@ -104,18 +111,23 @@ func initTurnServer(conf TurnConfig, auth func(username, real string, srcAddr ne
...
-			usersMap[kv[1]] = turn.GenerateAuthKey(kv[1], conf.Realm, kv[2])
-		}
-		if len(usersMap) == 0 {
-			log.Panicf("No turn auth provided")
-		}
-		auth = func(username string, realm string, srcAddr net.Addr) ([]byte, bool) {
-			if key, ok := usersMap[username]; ok {
-				return key, true
+		if conf.Auth.Secret != "" {
+			logger := logging.NewDefaultLeveledLoggerForScope("lt-creds", logging.LogLevelTrace, os.Stdout)
```
tarrencev, line 115: We have to pass a logger here. Once we migrate `ion-sfu` to use pions logger interface we can pass our logger here

OrlandoCo: Same case for webrtc internals, going to update the ion logger to follow the interface

OrlandoCo (approved):

PR #397, opened 2021-01-29
Allow signalling server to restart without interrupting audio/video.

grahamking:
A grpc stream can now identify itself with a string id. Identified
streams do not close their peer connections when the grpc connection is
interrupted. A grpc stream that identifies as an existing peer will
reconnect to that peer. Peers are deleted when they go to ICE state
closed.

This allows the signalling server to restart and reconnect to ion-sfu
without interrupting user audio/video.

A grpc stream that does not identify behaves exactly as before.

--- Discussion

*Help needed*: If the exact version of protocol buffers matters (I think it might), could someone regenerate the protocol buffers and grpc files with the correct versions? The docker setup did not work for me.

*Bikeshed*: The `GetIdentity` / `SetIdentity` / etc feels a bit, well, Java. I'm happy to replace it with direct attribute access if you prefer that style.


tarrencev:
I would prefer to wait for datachannel negotiations then simplify the grpc interface. The signaling modules in here are intended to be simple examples, so it is better not to introduce this extra complexity

grahamking:
Fair enough. I'll use my fork for now and migrate over when datachannel negotiation arrives.

grahamking:
For anyone using this functionality, you'll need this commit /  bugfix too: https://github.com/grahamking/ion-sfu/commit/25f77f0576d8b783ab37224988751515f4c6ce69

PR #401, opened 2021-02-01
Feat/temporal

OrlandoCo:
#### Description

This PR add support/ fix for the following:

- SFU honors simulcast preferred initial layer
- SFU will adjust simulcast spatial/temporal layers 
- Temporal layer  support added for vp8 packet
- Add audio observer


config.toml
```diff
@@ -19,7 +19,15 @@ bestqualityfirst = true
...
 enabletemporallayer = false
-
+# Sets the audio level volume threshold.
+# Values from [0-127] where 0 is the loudest.
+# Audio levels are read from rtp extension header according to:
+# https://tools.ietf.org/html/rfc6464
+audiolevelthreshold = 40
+# Sets the interval in which the SFU will emit the audio level
+# events in [ms]
+# Values 500-5000
```
grahamking, line 29: Why limit the interval to 500-5000? I could easily see a case for <500ms, 300 and even 100 seem reasonable values. On the other end I can't see why anyone would want >5s, but why limit it, probably some strange use cases out there.

OrlandoCo: We can lower or remove the limits, Just not sure if something less than 25 sample would be useful for something.

So 50-5000 is better or just remove it?

pkg/sfu/audiolevel.go
```diff
@@ -0,0 +1,91 @@
...
+type audioStream struct {
+	id    string
+	sum   int
+	total int
+}
+
+type audioLevel struct {
+	sync.RWMutex
+	threshold uint8
+	streams   []*audioStream
```
grahamking, line 17: Should streams be a map? It would make `observe` and `removeStream` simpler (and faster). It might make `calc` a little more complex though.

config.toml
grahamking: Ideally the interval at which we send the "X is speaking" messages wouldn't affect the interval over which we calculate if someone is speaking or not.

I vote just remove it. Maybe if it's <100ms (or 300ms?) log a warning, just to cut down on "server is slow" bug reports?

pkg/buffer/buffer.go
```diff
@@ -329,12 +339,21 @@ func (b *Buffer) calc(pkt []byte, arrivalTime int64) {
...
 		}
 	}
 
+	if b.audioLevel {
+		if e := p.GetExtension(b.audioExt); e != nil && b.onAudioLevel != nil {
+			ext := rtp.AudioLevelExtension{}
+			if err := ext.Unmarshal(e); err == nil {
+				b.onAudioLevel(ext.Level)
+			}
+		}
```
grahamking, line 354: For GetExtension and Unmarshal could we log the error at debug level. That could be really helpful for debugging.

If there is an error I think we should call `onAudioLevel` with 127. Otherwise we might think a user is still making noise, when instead they're having network problems.

config.toml
OrlandoCo: well with this implementation it is affected, since the values are resetted on timer, since I dont want to keep more buffers to prevent many allocations.

pkg/sfu/audiolevel.go
```diff
@@ -0,0 +1,91 @@
...
+	for idx, s := range a.streams {
+		if s.total > 0 {
+			streamIDs[idx] = s.id
+		}
+		s.total = 0
+		s.sum = 0
+	}
+
+	return streamIDs
+}
```
grahamking, line 91: This will return all the streams for which we got a packet of noise since the last call, ordered by number of samples then by loudness. I don't think that will give session.go enough information to calculate if someone is speaking or not.

pkg/buffer/buffer.go
OrlandoCo: If we log the error, we are going to get crazy errors logging there, the only way we could get an error there if it's for some reason the ext map ID is wrong.

It's not neccesary to send 127 since if value is wrong is not going to get registered.

pkg/sfu/audiolevel.go
grahamking: I pulled out the audio parts of our code to a gist: https://gist.github.com/grahamking/4cd83cb22186d8a7e700918520dcd6be

OrlandoCo: The number of samples which are over the threshold, so this will return all stream which get over the threshold where the first is the one with more packets over the threshold, if there are many with the same amount of packets over the threshold it will be ordered aslo by the loudest

pkg/sfu/session.go
```diff
@@ -189,3 +195,35 @@ func (s *Session) Peers() map[string]*Peer {
...
+
+		for _, peer := range s.peers {
+			if ch, ok := peer.subscriber.channels[APIChannelLabel]; ok {
+				if err = ch.SendText(sl); err != nil {
+					log.Errorf("Sending audio levels to peer: %s, err: %v", peer.id, err)
+				}
+			}
+		}
+	}
+}
```
grahamking, line 230: There's more needed here right? At the moment it sends an array of stream ids, but there's no way for the client to know which ones are making noise.

OrlandoCo: there is only returned streams with noise ordered:

```golang
	if dBov <= a.threshold {
				as.sum += int(dBov)
				as.total++
			}
```

grahamking (approved):

config.toml
```diff
@@ -19,7 +19,19 @@ bestqualityfirst = true
...
 enabletemporallayer = false
-
+# Sets the audio level volume threshold.
+# Values from [0-127] where 0 is the loudest.
+# Audio levels are read from rtp extension header according to:
+# https://tools.ietf.org/html/rfc6464
+audiolevelthreshold = 40
+# Sets the interval in which the SFU will emit the audio level
+# events in [ms]
+audiolevelinterval=1000
```
tarrencev, line 29: does this mean a client will only receive audio updates every second? is it possible to instead trigger based on a new active speaker?

grahamking: It will only send the message if the speaker(s) has changed, so you could set it to quite a short interval. I guess the comment should now say `in which the SFU will **check** the audio level`.

OrlandoCo: the problem is to identify when is there an active speaker and not a noise, the audiolevel will not send an event if there is no changes in the speakers, but it will wait the interval to check the loudest peer. so fi Peer A is speaking for 1 minute there will be only one speaker event in the DC API

config.toml
```diff
@@ -19,7 +19,19 @@ bestqualityfirst = true
...
 # enable only for testing.
 enabletemporallayer = false
-
+# Sets the audio level volume threshold.
+# Values from [0-127] where 0 is the loudest.
+# Audio levels are read from rtp extension header according to:
+# https://tools.ietf.org/html/rfc6464
+audiolevelthreshold = 40
+# Sets the interval in which the SFU will emit the audio level
+# events in [ms]
```
tarrencev, line 28:
```suggestion
# Sets the interval in which the SFU will check the audio level
# in [ms]. If the active speaker has changed, the sfu will
# emit an event to clients.
```

PR #410, opened 2021-02-05
Remove the offer sdp from peer.Join. fix #409

### LINKED ISSUE #409: Remove the offer parameter from peer.Join.
Separate the offer sdp parameter in the peer.Join method, so that it can be better in the signaling design.

before:
```go
                       answer, err := peer.Join(payload.Join.Sid, offer)
                        if err != nil {
                                switch err {
                                case sfu.ErrTransportExists:
                               ...
                        }
```
after:
```go
                       err = peer.Join(payload.Join.Sid)
                        if err != nil {
                                switch err {
                                case sfu.ErrTransportExists:
                                  return sig.SendReply(JoinReply{Success: false, Error: sfu.ErrTransportExists })
                                case sfu.RoomIsFull:
                                  return sig.SendReply(JoinReply{Success: false, Error: sfu.RoomIsFull })
                                 ...
                        }
                        return sig.SendReply(JoinReply{Success: true, Error: nil })

                       ......
                       // recv description.
                       answer, err := peer.Answer(offer)
                       if err != nil {
                               return status.Errorf(codes.Internal, fmt.Sprintf("answer error: %v", err))
                       }
```

When designing signaling, we need to distinguish the offer/answer/trickle negotiation process of webrtc from join/leave.

- The join method is a one-time one-way (C -> S) call,
- offer/answer/trickle is executed multiple times and is a two-way call
    C <----offer/answer/trickle---> S

The following is the new ion signaling, used for rtc negotiation between nodes or node and client

```protobuf
// Used for standard signaling between any nodes
// sfu<->sfu, sfu<->avp, client<->sfu, sfu<->sip-gw
service RTC {
    rpc Signal(stream Signalling) returns (stream Signalling) {}
}

message JoinRequest {
    string sid = 1;
    string uid = 2;
}

message JoinReply {
    boolean success = 1;
    string error = 2; // room is full ?
}

message Join {
    oneof payload {
           JoinRequest req = 1;
           JoinReply reply = 2;
    }
}

message Signalling {
    oneof payload {
        Join join = 1;
        // negotiation
        Description description = 2;
        Trickle trickle = 3;
        Error error = 4;
    }
}
```

cloudwebrtc:
Separate the offer sdp parameter in the peer.Join method.
This allows for a more flexible signal design. The current modification does not affect the existing ion-sfu signaling but separates the first offer SDP from the join method. When the join fails, we don’t have to operate pc or create an offer on the client-side.

before:
```go
                       answer, err := peer.Join(payload.Join.Sid, offer)
                        if err != nil {
                                switch err {
                                case sfu.ErrTransportExists:
                               ...
                        }
```
after:
```go
                       err = peer.Join(payload.Join.Sid)
                        if err != nil {
                                switch err {
                                case sfu.ErrTransportExists:
                                  return sig.SendReply(JoinReply{Success: false, Error: sfu.ErrTransportExists })
                                case sfu.RoomIsFull:
                                  return sig.SendReply(JoinReply{Success: false, Error: sfu.RoomIsFull })
                                 ...
                        }
                        return sig.SendReply(JoinReply{Success: true, Error: nil })

                       ......

                       // recv description.
                       answer, err := peer.Answer(offer)
                       if err != nil {
                               return status.Errorf(codes.Internal, fmt.Sprintf("answer error: %v", err))
                       }
```

#### Reference issue
Fixes #409

### DIFF
--- a/cmd/signal/grpc/server/server.go
+++ b/cmd/signal/grpc/server/server.go
@@ -148,7 +148,7 @@ func (s *SFUServer) Signal(stream pb.SFU_SignalServer) error {
 				}
 			}
 
-			answer, err := peer.Join(payload.Join.Sid, offer)
+			err = peer.Join(payload.Join.Sid)
 			if err != nil {
 				switch err {
 				case sfu.ErrTransportExists:
@@ -168,6 +168,11 @@ func (s *SFUServer) Signal(stream pb.SFU_SignalServer) error {
 				default:
 					return status.Errorf(codes.Unknown, err.Error())
 				}
+			}
+
+			answer, err := peer.Answer(offer)
+			if err != nil {
+				return status.Errorf(codes.Internal, fmt.Sprintf("answer error: %v", err))
 			}
 
 			marshalled, err := json.Marshal(answer)
--- a/cmd/signal/json-rpc/server/server.go
+++ b/cmd/signal/json-rpc/server/server.go
@@ -70,7 +70,13 @@ func (p *JSONSignal) Handle(ctx context.Context, conn *jsonrpc2.Conn, req *jsonr
 			}
 		}
 
-		answer, err := p.Join(join.Sid, join.Offer)
+		err = p.Join(join.Sid)
+		if err != nil {
+			replyError(err)
+			break
+		}
+
+		answer, err := p.Answer(join.Offer)
 		if err != nil {
 			replyError(err)
 			break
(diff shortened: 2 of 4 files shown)


billylindeman:
It was originally this way, and @tarrencev added it to the join to streamline the process awhile back.  thoughts @tarrencev

tarrencev:
i'm ok with these changes

adwpc:
LGTM

adwpc (approved):

PR #421, opened 2021-02-10
Adding logger package based on zerolog that implements logr

mikhno-s:
#### Description

This PR changes the ion-log module to package inside this repo that use zerolog to implement logr interface.


adwpc:
Thanks!
What are the advantages of logr

mikhno-s:
There is a couple advantages:
1) Logr provides interface for structured logging (easy to query), no need to printf with custom format.
2) It's simple - just need to create logger instance and pass it into the modules and packages
3) V-levels (-v, -vv, ...) instead of "string"-levels. However - it's still can be used along with "string" levels like "trace", "debug"
4) When creating in a proper way (one logger instance that will be called from other methods, or one logger instance per module/lib with different settings) - concurrently safe.  

However, current implementation support old way to make logs - initiate logger one time in main func.

pkg/logger/zerologr.go
```diff
@@ -0,0 +1,226 @@
...
+)
+
+var (
+	log Zerologr
+	mu  sync.RWMutex
+)
+
+// this prevents from using non-initialised logger
+func init() {
+	fixByFile := []string{"asm_amd64.s", "proc.go", "zerologr.go"}
```
bshihr, line 43: I know this was code you inherited but I'm curious if you know what the reason for these changes were.

pkg/logger/zerologr.go
```diff
@@ -0,0 +1,226 @@
...
+	logr.Logger
+	Infof(string, ...interface{})
+	Errorf(string, ...interface{})
+	Panicf(string, ...interface{})
+}
+
+// Config seems like isn't somewhere used, so just leave it here to don't let old code fail
+type Config struct {
+	Level string `mapstructure:"level"`
+}
```
bshihr, line 58: Let's get rid of this as silent failures / no-ops are probably worse than build failures.

cmd/signal/allrpc/main.go
```diff
@@ -5,8 +5,8 @@ import (
 	"fmt"
 	"os"
 
-	log "github.com/pion/ion-log"
 	"github.com/pion/ion-sfu/cmd/signal/allrpc/server"
+	log "github.com/pion/ion-sfu/pkg/logger"
```
bshihr, line 9: An important aspect of this effort is to change ion-sfu's approach of using a global hard-coded logger.  Ultimately, only "main" packages should ever call "github.com/pion/ion-sfu/pkg/logger" and all library code (non-main packages) should be using injected logr.Logger instances.

Is this planned for a follow-up change?

pkg/logger/zerologr.go
```diff
@@ -0,0 +1,226 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// package zerologr defines an implementation of the github.com/go-logr/logr
+// interfaces built on top of zerolog (github.com/rs/zerolog).
+
+package logger
```
bshihr, line 19:
```suggestion
// package logger defines a default implementation of the github.com/go-logr/logr
// interfaces built on top of zerolog (github.com/rs/zerolog) and is the default
// implementation for ion-sfu released binaries.
package logger
```

pkg/logger/zerologr.go
```diff
@@ -0,0 +1,226 @@
...
+	timeFormat     = "2006-01-02 15:04:05.000"
+)
+
+var (
+	log Zerologr
+	mu  sync.RWMutex
+)
+
+// this prevents from using non-initialised logger
+func init() {
```
bshihr, line 42: Let's avoid global side-effects.  This calls Init which calls zerolog which will result in a bad time for any client code that imports this module and has already done their own zerolog configuration.

pkg/logger/zerologr.go
```diff
@@ -0,0 +1,226 @@
...
+	log.Info(fmt.Sprintf(format, v...))
+}
+
+// Errorf logs a formatted error level log to the console
+func (l logger) Errorf(format string, v ...interface{}) {
+	l.Error(nil, fmt.Sprintf(format, v...))
+}
+
+func Errorf(format string, v ...interface{}) {
+	mu.RLock()
```
bshihr, line 197: We don't need any locking in this package.  Initialization code need not be thread-safe and logging objects are already safe.

pkg/logger/zerologr.go
mikhno-s: In current state it works when `Debugf()` is called from some function in buffer.go which is called from some test function in buffer_test.go. In this case logger isn't initialised and 
1) It prevent from null pointer exception
2) fixing function name and file name in log.  ( `...[zerologr.go] [Debugf]...` )

But if we do everything in a right way, and logger will be initialised in pkg itself (or even in test) - we don't need this hacky workaround at all.

mikhno-s:
@bshihr Okay, but what about tests packages? Should test code import the logr implementation?

bshihr:
> @bshihr Okay, but what about tests packages? Should test code import the logr implementation?

Seems reasonable to me.

mikhno-s:
I was trying to implement different V-levels, like described here - https://github.com/go-logr/logr#what-if-my-v-levels-differ-between-libraries. That works almost fine. 
But then I realised, that V level for logger should be additive. And it breaks my implementation. 
Still working on it.

pkg/sfu/sfu.go
```diff
@@ -151,6 +155,14 @@ func init() {
 
 // NewSFU creates a new sfu instance
 func NewSFU(c Config) *SFU {
+	// Init loggers
+	if c.Logger == nil {
+		println("Please pass logr instance implementation in SFU Config. You can use https://github.com/pion/ion-sfu/blob/5e6f7a3f58f0411d65ef4b31e8b6d58b6afd16da/pkg/logger/zerologr.go")
+		os.Exit(1)
+	}
+	defaultLogger = c.Logger
+	debugLogger = c.Logger.V(1)
```
bshihr, line 164: We should avoid globals like this.  Inject c.Logger into whatever needs it (e.g. NewBufferFactory below should take a logger argument).

bshihr (commented):
Sorry about the delay.  I think we're getting close but I think we may want to simplify (see comment about dependency injection vs package "Logger" globals.  My intuition is that logging is so messy right now that dependency injection will be a really big and breaking change so we can simplify further and just establish Logger instances.  CC @tarrencev

pkg/buffer/buffer.go
```diff
@@ -118,6 +123,14 @@ func NewBuffer(ssrc uint32, vp, ap *sync.Pool) *Buffer {
 func (b *Buffer) Bind(params webrtc.RTPParameters, o Options) {
 	b.Lock()
 	defer b.Unlock()
+
+	if o.Logger == nil {
+		println("Please pass logr instance implementation in Buffer Config. You can use https://github.com/pion/ion-sfu/blob/5e6f7a3f58f0411d65ef4b31e8b6d58b6afd16da/pkg/logger/zerologr.go")
+		os.Exit(1)
+	}
+	defaultLogger = o.Logger
+	debugLogger = o.Logger.V(1)
```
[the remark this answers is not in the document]
mikhno-s: I am not sure, that different log formats are good solution. It might be painful enough for the people who use sfu and want to process somehow their logs.

So, we can do the trick with public Logger instance, but ask users to reinit it with own logr if they want to have same log format everywhere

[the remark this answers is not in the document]
bshihr: > that different log formats are good solution

Can you elaborate on what you mean?  I don't think either suggestion changes how easy or hard it is to process logging output?  I expect users will want to set their own loggers (doable simply via `pkgfoo.Logger = MyOwnLogger()` somewhere in their binary's main pkg)?  That is certainly, for us, what is motivating the change; we have no interest in the logging implemented in the package here.

pkg/logger/zerologr_test.go
```diff
@@ -19,17 +18,41 @@ func captureOutput(f func()) string {
...
 
-	output = captureOutput(func() {
-		logger.V(2).Info("This is debug log")
-	})
-	assert.Equal(t, "", output) // Must be empty
+	// output = captureOutput(func() {
+	// 	logger.V(2).Info("This is debug log")
+	// })
+	// assert.Equal(t, "", output) // Must be empty
 }
```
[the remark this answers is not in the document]
mikhno-s: check the updated tests

pkg/buffer/buffer.go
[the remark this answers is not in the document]
mikhno-s: Sorry, are NOT good solution, I mean. Never mind, we will show the user Warning and create DiscardLogger if user did not provide own one.

I made defaultLogger global in buffer package, for example, because like [here](https://github.com/pion/ion-sfu/blob/aa347bbaa0151503bbeded9d04989e8daf797785/pkg/buffer/helpers.go#L168) - isH264Keyframe - it's just a function, not a method that can use logger as a property.

What is the best way to deal with cases like that? WDYT? @bshihr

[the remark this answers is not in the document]
bshihr: > I made defaultLogger global in buffer package, for example, because like here - isH264Keyframe - it's just a function, not a method that can use logger as a property.

Yea that's why I suggested the second option of just upgrading it to a public package variable (`Logger`) with a silent default (DiscardLogger) and using it everywhere which would require end-users to assign their desired values.

pkg/buffer/buffer.go
```diff
@@ -22,6 +24,8 @@ const (
 	reportDelta = 1e9
 )
 
+var defaultLogger, debugLogger logr.Logger
```
bshihr, line 27:
```suggestion
var Logger logr.Logger = new(logr.DiscardLogger)
```

And update the binaries/main packages to explictly set logging instances

```
func setLoggers(/* some parameters via cmd line or similar */) {
    buffer.Logger = log.New()
    ...
}
```

pkg/buffer/buffer.go
```diff
@@ -102,6 +106,7 @@ type Stats struct {
 // BufferOptions provides configuration options for the buffer
 type Options struct {
 	MaxBitRate uint64
+	Logger     logr.Logger
```
bshihr, line 109: And just get rid of all these embedded ones.

pkg/buffer/buffer.go
```diff
@@ -145,13 +158,13 @@ func (b *Buffer) Bind(params webrtc.RTPParameters, o Options) {
 		for _, fb := range codec.RTCPFeedback {
 			switch fb.Type {
 			case webrtc.TypeRTCPFBGoogREMB:
-				log.Debugf("Setting feedback %s", webrtc.TypeRTCPFBGoogREMB)
+				debugLogger.Info("Setting feedback", "type", "webrtc.TypeRTCPFBGoogREMB")
```
bshihr, line 161:
```suggestion
				Logger.V(1).Info("Setting feedback", "type", "webrtc.TypeRTCPFBGoogREMB")
```

Here and elsewhere, just consume the package global logger.

mikhno-s: Will do the same for sfu

pkg/logger/zerologr.go
```diff
@@ -149,3 +179,8 @@ func (l logger) WithValues(kvList ...interface{}) logr.Logger {
 	new.values = append(new.values, kvList...)
 	return new
 }
+
+// Verify that logger implements logr.Logger.
+var _ logr.Logger = logger{}
+
+// var _ logr.CallDepthLogger = logger{}
```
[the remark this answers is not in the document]
mikhno-s: Done

pkg/buffer/buffer.go
```diff
@@ -118,6 +123,14 @@ func NewBuffer(ssrc uint32, vp, ap *sync.Pool) *Buffer {
 func (b *Buffer) Bind(params webrtc.RTPParameters, o Options) {
 	b.Lock()
 	defer b.Unlock()
+
+	if o.Logger == nil {
+		println("Please pass logr instance implementation in Buffer Config. You can use https://github.com/pion/ion-sfu/blob/5e6f7a3f58f0411d65ef4b31e8b6d58b6afd16da/pkg/logger/zerologr.go")
```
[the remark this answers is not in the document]
mikhno-s: Done

pkg/buffer/buffer.go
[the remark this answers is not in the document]
mikhno-s: Done

pkg/sfu/sfu.go
mikhno-s: Done

mikhno-s: Added.

pkg/buffer/buffer.go
mikhno-s: Removed

pkg/buffer/buffer.go
mikhno-s: Updated. But only for sfu, because in cmd/ is no explicit buffer package using

pkg/buffer/buffer.go
mikhno-s: Added

mikhno-s:
@bshihr  Could you please review after the changes?

pkg/buffer/buffer_test.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/pion/rtcp"
 
+	log "github.com/pion/ion-sfu/pkg/logger"
```
bshihr, line 9: nit: Avoid renaming unless there's a collision.  I don't see one?

bshihr (commented):
LGTM! Thank you for shepherding this through.

One naming nit.

bshihr (approved):

pkg/buffer/buffer_test.go
mikhno-s: Yes, just has been removed. As well as in sfu test .

tarrencev (approved):
awesome work! thank you @mikhno-s @bshi

PR #423, opened 2021-02-12
feat(join with uid)

### LINKED ISSUE #422: (feat) Join with uid
Add a uid in Join, this will allow user:
1 use uid they prefer
2 trace log from FE to BE

https://github.com/pion/ion-sdk-js/issues/141

https://github.com/pion/ion-sfu/blob/116cc2c09d60b95e76ea5956f6f5e7add4054a21/cmd/signal/grpc/proto/sfu.proto#L32
https://github.com/pion/ion-sfu/blob/116cc2c09d60b95e76ea5956f6f5e7add4054a21/cmd/signal/json-rpc/server/server.go#L73

adwpc:
#### Description
1 use uid they prefer
2 trace log from FE to BE

#### Reference issue
Fixes #422 #91


pkg/sfu/peer.go
```diff
@@ -57,26 +57,28 @@ func NewPeer(provider SessionProvider) *Peer {
 }
```
cloudwebrtc, line 58: Do we need to export p.id here?
```go
func (p *Peer) ID() string {
    return p.id
}

```

cloudwebrtc (approved):

PR #427, opened 2021-02-17
add queriable metadata and stream id to peer

nkonev:
#### Description

OpenVidu has clientData field https://docs.openvidu.io/en/2.16.0/reference-docs/REST-API/#the-connection-object https://docs.openvidu.io/en/2.16.0/tutorials/openvidu-insecure-vue/#finally-connect-to-the-session-and-publish-your-webcam

It is used for setting arbitrary data, for example userName. I would add same field to Peer structure in order to query it later (see rough examples of usage below).

Also I need to add streamId field in order to map incoming stream with some user:
```js
                this.clientLocal.ontrack = (track, stream) => {
                    console.debug("got track", track.id, "for stream", stream.id);
                    axios.get("/users?session=chat1").then((d)=>{drawUserNames(d.data, stream.id)})
```

Query session peers - somewhere in main.go
```
type PeerDto struct {
	Id string `json:"id"`
	Metadata string `json:"metadata"`
	StreamId string `json:"streamId"`
}
...
	http.Handle("/users", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		sessionId := r.URL.Query().Get("session")
		session, _ := s.GetSession(sessionId)
		arr := []PeerDto{}
		for _, peer := range session.Peers() {
			arr = append(arr, PeerDto{Id: peer.Id(), StreamId: peer.StreamId(), Metadata: peer.Metadata()})
		}
		if marshal, err := json.Marshal(arr); err != nil {
			log.Errorf("Error during marshalling")
			w.WriteHeader(500)
		} else {
			w.Write(marshal)
		}
	}))
```


nkonev:
golint failed due [to](https://github.com/pion/ion-sfu/pull/427/checks?check_run_id=1917633088#step:4:9124)
```
  Error: struct field `Uid` should be `UID` (golint)
9125
```

pkg/sfu/publisher.go
```diff
@@ -46,6 +47,7 @@ func NewPublisher(session *Session, id string, cfg WebRTCTransportConfig) (*Publ
 
 	pc.OnTrack(func(track *webrtc.TrackRemote, receiver *webrtc.RTPReceiver) {
 		log.Debugf("Peer %s got remote track id: %s mediaSSRC: %d rid :%s streamID: %s", p.id, track.ID(), track.SSRC(), track.RID(), track.StreamID())
+		p.streamID = track.StreamID()
```
tarrencev, line 50: a peer can have more than one stream associated with it

tarrencev (changes requested):
@nkonev thank you for your PR. It should be possible for this logic to exist in the signaling layer without pushing it into the sfu implementation. You can parse the sdp in the signaling layer and maintain the metadata there. I'd be happy to accept a PR that does that. As an example, you can parse track/stream info from the sdp like this:

```
parsed := sdp.SessionDescription{}
	if err := parsed.Unmarshal([]byte(desc.SDP)); err == nil {
		for _, md := range parsed.MediaDescriptions {
			if msid, ok := md.Attribute(sdp.AttrKeyMsid); ok {
				split := strings.Split(msid, " ")
				if len(split) != 2 {
					log.Error().Msgf("Invalid msid: %s", msid)
					continue
				}

				msid := split[0]
				trackID := split[1]

				// if transceiver is inactive, remove the its metadata
				if _, ok := md.Attribute(sdp.AttrKeyInactive); ok {
...
```

nkonev:
@tarrencev Thank you for response.
Indeed I didn't do comprehensive research where is the proper place to insert the metadata support :) 

It sounds that it's need to close this request and elaborate other somewhere [here](https://github.com/pion/webrtc).

PR #438, opened 2021-02-20
feat(buffer): Remove channel

OrlandoCo:
#### Description

Remove packet chan from buffer


adwpc:
* load-testing: 20 clients 
* There is no chanrecv cpu cost
image
image
image
image

adwpc (approved):
LGTM!

PR #444, opened 2021-02-27
Added allrpc Dockerfile

### LINKED ISSUE #440: Dockerfile for AllRPC
## Summary

The `AllRPC` version of `ion-sfu` currently has no Dockerfile. It would be useful to create a Dockerfile for `AllRPC` and build a version for every release comparable to what is happening for `json-rpc` and `grpc`.

## Motivation

The `json-rpc` and `grpc` versions can easily be started using Docker, but `allrpc` currently cannot be started using Docker. This would be a great feature for further automating the `ion-sfu` deployment.

## Implementation

`AllRPC` has multiple parameters that can be defined to customize the setup. These parameters are currently defined using CLI parameters and currently cannot be defined in the TOML config file. Therefore the following changes need to be implemented for adding the Dockerfile.

1. Implement the [AllRPC parameters](https://github.com/pion/ion-sfu/blob/master/cmd/signal/allrpc/main.go#L76-L79) into the [SFU config](https://github.com/pion/ion-sfu/blob/master/pkg/sfu/sfu.go#L46) 
2. The parameters need to be read from the config file instead of CLI parameters (can also be overwritten using CLI parameters)
3. The Dockerfile needs to be implemented similar to the `ion-sfu>json-rpc` Dockerfile

TannerGabriel:
#### Description

Dockerfile for the `allrpc` version of `ion-sfu`. Command-line arguments for configuration can be provided using environment variables:

```bash
docker run -p 7000:7000 -p 8000:8000 -e gaddr=0.0.0.0:8000 -e jaddr=0.0.0.0:7000  allrpc
```

#### Reference issue

Fixes #440

### DIFF
--- a/cmd/signal/allrpc/Dockerfile
+++ b/cmd/signal/allrpc/Dockerfile
@@ -1,0 +1,24 @@
+FROM golang:1.14.14-stretch
+
+ENV GO111MODULE=on
+
+WORKDIR $GOPATH/src/github.com/pion/ion-sfu
+
+COPY go.mod go.sum ./
+RUN cd $GOPATH/src/github.com/pion/ion-sfu && go mod download
+
+COPY pkg/ $GOPATH/src/github.com/pion/ion-sfu/pkg
+COPY cmd/ $GOPATH/src/github.com/pion/ion-sfu/cmd
+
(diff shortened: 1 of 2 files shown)


adwpc:
LGTM

PR #445, opened 2021-02-28
Mute should not cause timestamps to freeze

tarrencev:
Timestamps needs to continue going forward even when muted.
NTP needs to remain in sync with progressing timestamps.


adwpc:
Do we need to add `d.reBaseTs.set(true)` in Mute()?

PR #447, opened 2021-03-01
fix nack handling for downtracks, enable external bufferfactory

davidzhao:
#### Description
I was experiencing freezes in my streams, and it turned out that browser has not been able to use the packets that were retransmitted due to incorrect SSRC and payloadType. This PR addresses it.

It also enables third-party bufferFactory, so I could go back to using ion-sfu instead of my forked copy for my project.


adwpc:
LGTM
@OrlandoCo @tarrencev 
How about the bufferFactory?

PR #448, opened 2021-03-01
Adding allrpc Github workflow

TannerGabriel:
#### Description

Github workflow for automatically pushing allrpc Docker image similar to what already happens with [jsonrpc](https://github.com/pion/ion-sfu/blob/master/.github/workflows/docker-jsonrpc.yaml) and [grpc](https://github.com/pion/ion-sfu/blob/master/.github/workflows/docker-grpc.yaml).

Docker image will then be available as `pionwebrtc/ion-sfu:latest-allrpc` on Dockerhub.

### DIFF
--- a/.github/workflows/docker-allrpc.yaml
+++ b/.github/workflows/docker-allrpc.yaml
@@ -1,0 +1,34 @@
+name: allrpc docker
+on:
+  push:
+    branches:
+    - master
+  release:
+    types: [published]
+  pull_request:
+    branches:
+    - master
+jobs:
+  build:
+    name: build and push
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+    steps:
+      - uses: actions/checkout@v2
+    
+      - name: build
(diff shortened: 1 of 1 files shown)


adwpc (approved):
LGTM

.github/workflows/docker-allrpc.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: tag
+        if: github.event_name == 'release'
+        run: docker tag pionwebrtc/ion-sfu:latest-allrpc pionwebrtc/ion-sfu:"$TAG"-allrpc
+        env:
+          TAG: ${{ github.event.release.tag_name }}
+
+      - name: push
+        if: github.event_name == 'release'
+        run: docker push pionwebrtc/ion-sfu
```
tarrencev, line 34: there were some changes to docker it seems. i believe we need to explicitly push image tags now

https://github.com/pion/ion-sfu/blob/master/.github/workflows/docker-grpc.yaml#L34

TannerGabriel: Ok did not know about that. It should be fixed now!

PR #462, opened 2021-03-12
fix: pubsubtest compose uses allrpc

leewardbound:
I was just made aware we provide this compose file in `pubsubtest` folder; if we use allrpc instead of jsonrpc, this makes a great entrypoint for introducing people to `ion-sdk-go`, `ion-avp` or `ion-app-flutter` (because jsonrpc is great for a simple pubsubtest, but everything else uses grpc)


adwpc (approved):
LGTM

PR #471, opened 2021-03-15
feat(sfu): add join config to peers

OrlandoCo:
#### Description

Add join configuration to peers that will allow to disable publishing or subscribing, this will allow other services or relaying to not open useless peerconnections.

The configuration is optional hence backwards compatible.

### DIFF
--- a/pkg/sfu/peer.go
+++ b/pkg/sfu/peer.go
@@ -24,6 +24,14 @@ var (
 	ErrOfferIgnored = errors.New("offered ignored")
 )
 
+// JoinConfig allow adding more control to the peers joining a session.
+type JoinConfig struct {
+	// If true the peer will not be allowed to publish tracks to session.
+	NoPublish bool
+	// If true the peer will not be allowed to subscribe to other peers in session.
+	NoSubscribe bool
+}
+
 // SessionProvider provides the session to the sfu.Peer{}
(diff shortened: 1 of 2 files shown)


tarrencev (approved):

pkg/sfu/peer.go
```diff
@@ -74,80 +87,85 @@ func (p *Peer) Join(sid, uid string) error {
...
+		p.publisher, err = NewPublisher(p.session, uid, cfg)
+		if err != nil {
+			return fmt.Errorf("error creating transport: %v", err)
 		}
-
-		if p.OnIceCandidate != nil && !p.closed.get() {
-			json := c.ToJSON()
-			p.OnIceCandidate(&json, publisher)
+		for _, dc := range p.session.datachannels {
+			if err := p.subscriber.AddDatachannel(p, dc); err != nil {
```
SnaiCrys, line 138: As far as I understand, there will always be an error in the case of `conf.NoSubscribe == true`

OrlandoCo: You are right!

SnaiCrys: In general, it turned out that with this new conf, the "two pc system" (https://github.com/pion/ion-sdk-js/commit/6dff082093893ae7befc1317e78cf05867faa464) in the ion-sdk-js changed a lot.

PR #472, opened 2021-03-15
signal/grpc: ability to pprof the server

gobwas:
#### Description
This commit adds an ability to pprof the server when running by providing `-paddr` flag with `host:port` addr string.

### DIFF
--- a/cmd/signal/grpc/main.go
+++ b/cmd/signal/grpc/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	_ "net/http/pprof"
 	"os"
 
 	"github.com/pion/ion-sfu/pkg/middlewares/datachannel"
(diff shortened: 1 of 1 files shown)


adwpc (approved):
LGTM

cmd/signal/grpc/main.go
```diff
@@ -97,8 +101,14 @@ func parse() bool {
 	flag.StringVar(&addr, "a", ":50051", "address to use")
 	flag.StringVar(&metricsAddr, "m", ":8100", "merics to use")
 	flag.IntVar(&verbosityLevel, "v", -1, "verbosity level, higher value - more logs")
+	flag.StringVar(&paddr, "paddr", "", "pprof listening address")
 	help := flag.Bool("h", false, "help info")
 	flag.Parse()
+
+	if paddr == "" {
+		paddr = getEnv("paddr")
```
adwpc, line 109: Maybe we can use "PPROF_ADDR":)

gobwas: I used it to be in sync with all-rpc version of the server:
https://github.com/pion/ion-sfu/blob/191ae903e730d8e36169d02faf2b33fb416c69e9/cmd/signal/allrpc/main.go#L107

grahamking:
LGTM also

PR #477, opened 2021-03-18
Simulcast: Allow layer selection for publisher

dreamerns:
#### Description
* Allow publisher to select which layer(s) should be enabled. 
* Publisher can activate/deactivate layer with encoding params of the sender, however, there is no easy way to detect if a layer is enabled or not on sfu side. I added another message to middleware that will allow publishers to select which layer(s) are currently active on the sender side. It updates all the down tracks accordingly. 
* Message is then fanned out to all subscribers, so they are aware which layer is active and which ones are available. 
* In order to send messages on 'ion-sfu' channel (besides audio levels), I added a little bit more robust messaging type (similar to json-rpc). This is a breaking change for the client, unfortunately, but it gives us more flexibility if any new message type needs to be added in the future.

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/sfu/peer.go
+++ b/pkg/sfu/peer.go
@@ -48,6 +48,11 @@ type SessionProvider interface {
 	GetSession(sid string) (Session, WebRTCTransportConfig)
 }
 
+type ChannelAPIMessage struct {
+	Method string      `json:"method"`
+	Params interface{} `json:"params,omitempty"`
+}
+
 // PeerLocal represents a pair peer connection
 type PeerLocal struct {
 	sync.Mutex
@@ -237,6 +242,22 @@ func (p *PeerLocal) Trickle(candidate webrtc.ICECandidateInit, target int) error
 		if err := p.subscriber.AddICECandidate(candidate); err != nil {
 			return fmt.Errorf("setting ice candidate: %w", err)
 		}
+	}
+	return nil
+}
+
+func (p *Peer) SendAPIChannelMessage(msg *[]byte) error {
+	if p.subscriber == nil {
+		return fmt.Errorf("No subscriber for this peer")
+	}
+	dc := p.subscriber.DataChannel(APIChannelLabel)
+
+	if dc == nil {
+		return fmt.Errorf("Data channel %s doesn't exist", APIChannelLabel)
+	}
+
+	if err := dc.SendText(string(*msg)); err != nil {
+		return fmt.Errorf("Failed to send message: %v", err)
 	}
 	return nil
 }
--- a/pkg/sfu/peer.go
+++ b/pkg/sfu/peer.go
@@ -30,6 +30,7 @@ type Peer interface {
 	Publisher() *Publisher
 	Subscriber() *Subscriber
 	Close() error
+	SendAPIChannelMessage(msg *[]byte) error
 }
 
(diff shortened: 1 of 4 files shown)


pkg/sfu/session.go
```diff
@@ -204,6 +204,15 @@ func (s *Session) OnClose(f func()) {
 	s.onCloseHandler = f
 }
 
+func (s *Session) SendMessage(method, params, origin string) {
```
billylindeman, line 207: Might be better named BroadcastAPIChannelMessage for the whole Session

pkg/sfu/peer.go
```diff
@@ -215,6 +226,25 @@ func (p *Peer) Trickle(candidate webrtc.ICECandidateInit, target int) error {
 	return nil
 }
 
+func (p *Peer) SendMessage(method, params string) error {
```
billylindeman, line 229: Maybe name SendAPIChannelMessage?

pkg/sfu/peer.go
```diff
@@ -30,6 +31,16 @@ type SessionProvider interface {
 	GetSession(sid string) (*Session, WebRTCTransportConfig)
 }
 
+const (
```
billylindeman, line 34: These might be better defined in the subscriberapi.go, not a strong opinion on this though just something to think about

billylindeman (commented):
awesome work

pkg/sfu/session.go
dreamerns: Agree. I wasn't too happy with my naming :-)

pkg/sfu/peer.go
dreamerns: I've been debating the same. Since the method name can be anything, we could put constants in the appropriate location (session and subscribeapi.go), that way middleware would be on one place, but on the other hand this way all available methods are in one place. Not a strong opinion either.

adwpc:
This is useful.

pkg/sfu/session.go
```diff
@@ -210,6 +209,15 @@ func (s *Session) setRelayedDatachannel(peerID string, datachannel *webrtc.DataC
 	})
 }
 
+func (s *Session) BroadcastAPIChannelMessage(msg ChannelAPIMessage, origin string) {
+	peers := s.Peers()
```
OrlandoCo, line 213: Can we make this a generic function? that accepts msg and channel?

dreamerns: Hey @OrlandoCo I addressed some of your comments from chat, making peer.SendAPIChannelMessage generic and preventing marshaling for every peer.

pkg/sfu/peer.go
```diff
@@ -235,6 +240,18 @@ func (p *Peer) Trickle(candidate webrtc.ICECandidateInit, target int) error {
 	return nil
 }
 
+func (p *Peer) SendAPIChannelMessage(msg *[]byte) error {
+	if p.subscriber == nil {
+		return fmt.Errorf("No subscriber for this peer")
+	}
+	dc := p.subscriber.DataChannel(APIChannelLabel)
```
billylindeman, line 247: We should nil check dc, even though it should never be nil

billylindeman (approved):

PR #487, opened 2021-03-23
Add `StartRemoteTracksMuted` join option

kevinmcconnell:
This adds a new join option, `StartRemoteTracksMuted`.  When set to true, the peer's remote tracks will start off in the muted state.  This makes it easier for a peer to selectively receive some remote tracks while ignoring others.

As part of this change, the initial enabled/disabled state of a `DownTrack` is now set on create, rather than `Bind`.  This is to help with cases where a client sets a track's state early on (such as in an `ontrack` handler).  Previously if the state was set before `Bind` it would be lost, which made it harder for clients to select their tracks reliably.

### DIFF
--- a/pkg/sfu/downtrack.go
+++ b/pkg/sfu/downtrack.go
@@ -42,13 +42,14 @@ type DownTrack struct {
 	spatialLayer  int32
 	temporalLayer int32
 
-	enabled  atomicBool
-	reSync   atomicBool
-	snOffset uint16
-	tsOffset uint32
-	lastSSRC uint32
-	lastSN   uint16
-	lastTS   uint32
+	startMuted bool
+	enabled    atomicBool
+	reSync     atomicBool
+	snOffset   uint16
+	tsOffset   uint32
+	lastSSRC   uint32
+	lastSN     uint16
+	lastTS     uint32
 
 	simulcast        simulcastTrackHelpers
 	maxSpatialLayer  int64
@@ -70,7 +71,7 @@ type DownTrack struct {
 }
 
 // NewDownTrack returns a DownTrack.
-func NewDownTrack(c webrtc.RTPCodecCapability, r Receiver, bf *buffer.Factory, peerID string, mt int) (*DownTrack, error) {
+func NewDownTrack(c webrtc.RTPCodecCapability, r Receiver, bf *buffer.Factory, peerID string, mt int, startMuted bool) (*DownTrack, error) {
 	return &DownTrack{
 		id:            r.TrackID(),
 		peerID:        peerID,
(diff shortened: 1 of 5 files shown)


kevinmcconnell:
Just noticed the failing test 🤦  Will fix that now.

OrlandoCo:
@kevinmcconnell it's ok if I do some changes to this branch?

kevinmcconnell:
@OrlandoCo  yes, of course!  Feel free to change it however you like.

cloudwebrtc:
@kevinmcconnell  Thank you for your PR, this is helpful for me to upgrade the new SFU signaling,
https://github.com/pion/ion/blob/re-design-core-protocol/proto/rtc/rtc.proto

In the new sfu node, I need not to subscribe to any stream at startup, and then select the required publisher.Tracks() to generate DownTracks by updating the set command, and then trigger subcribe.Negotiate()

So I want to continue to push forward the update of this PR, @OrlandoCo @adwpc 

let me briefly describe:

There are two ways of working,
1. The subcriber receives the sdp negotiation from ion-sfu (including the m-line of all pubs track) at any time, and then controls the stream/track through mute/unmute

2. Change the name of `StartRemoteTracksMuted` to `NoAutoSubcribe`, which means that after Peer join, SDP does not contain any pubs tracks. At this point, a Subscription command is required to carry trackIds to subscribe to a specific stream. The process looks as follows
```go

subscription := payload.UpdateSettings.GetSubcription()
enabled := subscription.GetSubscribe()

for _, trackId := range subscription.TrackIds {

	if enabled {
                 // Add down tracks
		for _, p := range peer.Session().Peers() {
			if p.ID() != peer.ID() {
				for _, track := range p.Publisher().PublisherTracks() {
					log.Debugf("Remote.Track: %v", track.Track.ID())
					if track.Receiver.TrackID() == trackId {
						peer.Publisher().GetRouter().AddDownTrack(peer.Subscriber(), track.Receiver)
					}
				}
			}
		}
	} else {
               // Remove down tracks
		for streamID, downTracks := range peer.Subscriber().DownTracks() {
			for _, track := range downTracks {
				if track.ID() != trackId {
					peer.Subscriber().RemoveDownTrack(streamID, track)
				}
			}
		}
	}

}

peer.Subscriber().Negotiate()
```

adwpc:
This pr useful

cloudwebrtc:
Submit a new PR here https://github.com/pion/ion-sfu/pull/558

kevinmcconnell:
@cloudwebrtc glad you found this useful!  I think we can close this PR, if your #558 has superseded it?

(The new signalling looks very interesting, by the way!  I haven't had time recently to keep up with the changes, but it looks to be moving in a great direction.)

PR #488, opened 2021-03-23
Add join config to gRPC signalling server

kevinmcconnell:
Allows the new join config options to be passed to the gRPC signalling server.

(Note: this doesn't yet include the `StartRemoteTracksMuted` flag from #487.  If that PR seems ready to land soon, I can update this to include that flag too.)


adwpc:
Thanks all the same.
#529

PR #491, opened 2021-03-24
onMessage without out + add datachannel to session

decanus:
#### Description

This adds code we use in Soapbox, the ability to add a datachannel on a singular session along with being able to have `on Message` without needing to have output channels.


OrlandoCo:
Is this just to set a callback without getting the other Peers DC?

decanus:
@OrlandoCo yes, and then also to add a data channel to a singular session rather than have the SFU add it to all sessions.

OrlandoCo:
@decanus We can get datachannels from session, I think that by default we can disable the datachannels so you can get them if you want, to not add more configs, are you ok with that?

decanus:
@OrlandoCo can you give me an example how that works / would work?

PR #496, opened 2021-03-29
FAQ entries on how to use ion-sfu

grahamking:
All written by @leewardbound

Until now we didn't provide any direction on how to actually use ion-sfu. Now we do!

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,5 +1,45 @@
 # Frequenty Asked Questions
 
+### Can I / should I build on ion-sfu?
+ 
+Ion-sfu is considered "stable beta" and you can safely build on it. Many features are still under development, and the API is subject to change rapidly in future releases. If you want to build production-ready apps with ion-sfu, you need a strong grasp of WebRTC fundamentals and modern web development, and some experience with pion and golang is strongly recommended. You'll need to design your own Application Layer, responsible for transmitting signalling messages, validating client input, authenticating your users, administering room membership, and orchestrating scaling decisions.
+
+### Is ion-sfu actively maintained or supported? Can I get help?
+
+Ion-sfu actively developed and supported by an awesome community of volunteers! As an open source project in the pion community, the people who support this project are just the same volunteers who've made time to build and release it. You can join us in the gophers slack in #ion, we have a helpful and positive community and many people are always glad to talk about the project! Once you learn and use ion-sfu, you can help make the community better, too!
+
+### Which API should I use?
+    
+There are many strong opinions about the answer to this question!
+
+- Library API - Use ion-sfu as a library. Start by embedding the SFU struct. Use the grpc or jsonrpc examples as a guide.
+
+- JSONRPC - a websocket interface for bi-directional json messages; it is a simpler and lighter protocol. It is very easy to parse and handle the messages in any language. The internal javascript examples mostly use jsonrpc because it is easy to read and understand.
+
+- GRPC - uses bi-directional message streams and protobufs. The client code is automatically generated for consumers, with strong typing definitions. Typically, no manual message parsing is required.
+
+- "All RPC" - package is jsonrpc + grpc at the same time, for convenient testing between a cmd line publisher and a bowser subscriber (or vice-versa).
+
+**None of these** are designed for end-users to connect directly to them. Always build an Application Server, either by using ion-sfu as a library, or using the grpc API to talk to a local ion-sfu. (?? Is using internal GRPC "officially supported"?)
+
+### What do you mean by "Application Server"?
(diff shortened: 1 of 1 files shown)


grahamking:
@tarrencev @OrlandoCo Please edit the text as you see fit. This came out of a long discussion on Slack between @leewardbound , @billylindeman and myself.

FAQ.md
```diff
@@ -1,5 +1,45 @@
...
+
+- Library API - Use ion-sfu as a library. Start by embedding the SFU struct. Use the grpc or jsonrpc examples as a guide.
+
+- JSONRPC - a websocket interface for bi-directional json messages; it is a simpler and lighter protocol. It is very easy to parse and handle the messages in any language. The internal javascript examples mostly use jsonrpc because it is easy to read and understand.
+
+- GRPC - uses bi-directional message streams and protobufs. The client code is automatically generated for consumers, with strong typing definitions. Typically, no manual message parsing is required.
+
+- "All RPC" - package is jsonrpc + grpc at the same time, for convenient testing between a cmd line publisher and a bowser subscriber (or vice-versa).
+
+**None of these** are designed for end-users to connect directly to them. Always build an Application Server, either by using ion-sfu as a library, or using the grpc API to talk to a local ion-sfu. (?? Is using internal GRPC "officially supported"?)
```
leewardbound, line 23: Tarrence + Orlando, can you 2 confirm whether using the grpc API (internally, after validating SDPs in an upstream application server) should currently be considered a "supported" usecase? Can we safely advise people to use the grpc or jsonrpc docker images *internally* as long as input is sanitized?

tarrencev: I think we should move the included signaling to be "examples" and the batteries included deployment should be "ion"

tarrencev:
This is super awesome!

FAQ.md
leewardbound: Okay, I think that represents a major change needed to the above advice, we will have to reword it to avoid saying jsonrpc/grpc are official interfaces people should build upon, I will mark this as a Draft and we can try to adjust the verbiage this week

leewardbound (changes requested):
We will have to reword it to avoid saying jsonrpc/grpc are official interfaces people should build upon, and clarify they are examples that you can begin your own Application Server from.

https://github.com/pion/ion-sfu/pull/496/files#r603736974

FAQ.md
tarrencev: Sounds good. The reason i think this is the right direction is there are a million different ways to signal and it is very dependent on your goals. I would prefer to remove it as a source of distraction. Then it will hopefully be more clear that authors should extend their own implementation rather than try to generalize it to other use cases.

leewardbound: I took some time to think about it and I understand why it seems like the best option - and the GoLang API is undoubtedly the strongest one to support. It is also the hardest to use, and adds a new requirement that everyone write GoLang to use ion-sfu.

I have to say, I do see this as breaking support for most existing projects built on ion-sfu -- including mine. Granted, by moving the entrypoints to `examples/` you are leaving a backup option, but I feel like you are suggesting we deprecate "official" support for 2 APIs that most people have been building on. Tarrence, you yourself told me I should build a python client to talk to SFU over grpc... I know this means lots of re-thinking and re-structuring and re-work for me and many others.

I tried to maintain my own SFU wrapper in golang, but it was a source of constant frustration and bugs, I abandoned the code after wasting 4mo evenings and weekends on the effort, I much preferred to be able to write an API client in a language I am familiar with. Mainstream support for grpc and jsonrpc has been implied (at least since the early versions of the codebase), by both the name of the folder they live in and their repeat appearance in related SDK project examples, and further reinforced by the prevailing discussions in the slack channel. I'm sad nobody spoke up and said "don't build on this grpc entrypoint, it should not be an official entrypoint" much sooner.

I'm always happy to see ion-sfu become a better project, but as a maintainer who devotes most of his resources to helping teach newcomers and share knowledge, I expect I'll be dealing with frustrated and sad users. We'll be best off to add a section to the FAQ `"What happened to cmd/signal?"` at the very least. I'll prepare the necessary changes.

tarrencev: I understand you concern and I definitely don't think that we should kill it. We should still continue maintaining it in its current form. My main point is to move away from it being seen as a signaling layer that competes with i.e. janus or jitsi signaling. Our goal is to move almost everything to happen over webrtc (using datachannels) so the out of bound signaling will be very simple. Anything on top will be application specific. I know it is frustrating to have a moving target, but we are all learning together and make mistakes. I think we would change a lot about how we do signaling if we knew then what we know now.

PR #505, opened 2021-04-23
Fix issue with h264 layer switch in simulcast

dreamerns:
* When we are making receiver initiated layer switch in simulcast using preferLayer() and h264 codec, we usually get artifact in stream due to missing keyframe.
* Switching layer immidiatelly and sending pli (instead of waiting for keyframe) fixes the issue.

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/sfu/receiver.go
+++ b/pkg/sfu/receiver.go
@@ -345,12 +345,11 @@ func (w *WebRTCReceiver) writeRTP(layer int) {
 		w.locks[layer].Lock()
 
 		if w.isSimulcast && len(w.pendingTracks[layer]) > 0 {
-			if pkt.KeyFrame {
-				w.downTracks[layer] = append(w.downTracks[layer], w.pendingTracks[layer]...)
-				w.pendingTracks[layer] = w.pendingTracks[layer][:0]
-			} else {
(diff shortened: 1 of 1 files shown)


billylindeman (approved):

pkg/sfu/receiver.go
```diff
@@ -345,12 +345,11 @@ func (w *WebRTCReceiver) writeRTP(layer int) {
...
 		if w.isSimulcast && len(w.pendingTracks[layer]) > 0 {
-			if pkt.KeyFrame {
-				w.downTracks[layer] = append(w.downTracks[layer], w.pendingTracks[layer]...)
-				w.pendingTracks[layer] = w.pendingTracks[layer][:0]
-			} else {
+			if !pkt.KeyFrame {
 				w.SendRTCP(pli)
 			}
+			w.downTracks[layer] = append(w.downTracks[layer], w.pendingTracks[layer]...)
+			w.pendingTracks[layer] = w.pendingTracks[layer][:0]
```
OrlandoCo, line 352: We should fix the h264 keyframe detection, this would affect VP8 since it will change layers as soon as requested.

OrlandoCo (changes requested):

PR #508, opened 2021-04-28
Added NewSFUWithTurnAuth function

nkonev:
#### Reference issue
Added convenient function to create SFU with turn with auth callback


OrlandoCo:
hey @nkonev I think we can add the auth func in the config instead adding a custom method what do you think?

OrlandoCo (approved):

PR #524, opened 2021-05-12
fix(sfu): make peer dc msg method not only for api

OrlandoCo:
#### Description

just small change to not make interface function for api only.

### DIFF
--- a/pkg/sfu/peer.go
+++ b/pkg/sfu/peer.go
@@ -254,7 +254,7 @@ func (p *PeerLocal) SendDCMessage(label string, msg *[]byte) error {
 	dc := p.subscriber.DataChannel(label)
 
 	if dc == nil {
-		return fmt.Errorf("data channel %s doesn't exist", APIChannelLabel)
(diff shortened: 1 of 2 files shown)


pkg/sfu/peer.go
```diff
@@ -247,18 +247,18 @@ func (p *PeerLocal) Trickle(candidate webrtc.ICECandidateInit, target int) error
...
 	if p.subscriber == nil {
-		return fmt.Errorf("No subscriber for this peer")
+		return fmt.Errorf("no subscriber for this peer")
 	}
-	dc := p.subscriber.DataChannel(APIChannelLabel)
+	dc := p.subscriber.DataChannel(label)
 
 	if dc == nil {
-		return fmt.Errorf("Data channel %s doesn't exist", APIChannelLabel)
+		return fmt.Errorf("data channel %s doesn't exist", APIChannelLabel)
```
billylindeman, line 257: s/APIChannelLabel/label

;)

dreamerns (approved):
agree, this is better :-)

PR #530, opened 2021-05-17
feat(sfu): add webrtc timeouts config

panki:
#### Description

Allows configuring `webrtc.SettingEngine` timeouts. Default values don't work for all cases. For example, during debug session, it would be great to increase timeouts to be able to stop client application in the debugger and keep connections opened to be able to resume.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -70,6 +70,14 @@ mdns = true
 # nat1to1 = ["1.2.3.4"]
 # icelite = true
 
+[webrtc.timeouts]
+# The duration in [sec] without network activity before a ICE Agent is considered disconnected
+disconnected = 5
+# The duration in [sec] without network activity before a ICE Agent is considered failed after disconnected
+failed = 25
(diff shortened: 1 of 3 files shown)


adwpc:
Looks good, @OrlandoCo

OrlandoCo:
@panki this looks great, but since this changes the config file, can we check that if the values are 0 defaults them to some value and log a warn?

We got some issues before when config files were updated, people got many issues because they didt catch the change in the file

panki:
> @panki this looks great, but since this changes the config file, can we check that if the values are 0 defaults them to some value and log a warn?
> 
> We got some issues before when config files were updated, people got many issues because they didt catch the change in the file

@OrlandoCo this is a very good point, thank you, done

panki:
@OrlandoCo Do you have any guess why tests fails by timing out? locally everything is good.

OrlandoCo:
@panki I think is a ci related issue, since tests  are always successful on my local machine

PR #540, opened 2021-06-21
pli throttling

frostbyte73:
#### Description

Currently, the WebRTCReceiver throttles PLIs every 500ms, however if multiple are requested within 500ms only the first one is sent - this means if one client requests it 490ms after another, the second will get dropped. This update keeps the same throttling period but will send a pli at the end of the period if another was requested. 

It also adds the duration as a router config variable.


frostbyte73:
Changed this to a config var only - after some testing, realized the second one being dropped is what we want in most cases, since the clients spam plis until they get a full frame

PR #544, opened 2021-06-30
fix(sfu): remove race condition for session & peer

hn8:
#### Description
After network disruption, client may restart signaling with the same uid. It is possible that new AddPeer is called before RemovePeer in the exiting goroutine that delete the newly created one.

This fix also prevents duplicated NewSession in race condition. Also fixes GetSessions that the returned map is not thread-safe.


hn8:
GetSessions() is currently broken, map iteration would panic in race. Cloning map is more expensive than slice. Hope to hear feedback from the maintainers. Thanks

OrlandoCo:
@adwpc please check if the signature change breaks ION, this is the correct way to get the session since maps share the same inner references so is not thread safe.

adwpc:
LGTM，seems not break

adwpc (approved):

hn8:
Added 1 more commit for minor tweak of mutex for Subscriber.DataChannel and AudioObserver. 

Note: There are 2 identical method for get datachannel, and AddDataChannel vs AddDatachannel. I understand it is not a good idea to change public API of ion-sfu. But maybe rename AddDatachannel (lowercase c) method to InitDatachannel to reduce confusion in next version of API?

hn8:
@adwpc @OrlandoCo Would you please take a look if PR looks good to you? Thanks

OrlandoCo (approved):

adwpc (approved):

PR #546, opened 2021-07-03
refactor(sfu): optimize AudioObserver

hn8:
#### Description
Even in large session, list of active speakers is always short. Change to sort/clone smaller list, use map for streams, and reuse slice.

### DIFF
--- a/pkg/sfu/audioobserver.go
+++ b/pkg/sfu/audioobserver.go
@@ -12,8 +12,8 @@ type audioStream struct {
 }
 
 type AudioObserver struct {
-	sync.RWMutex
-	streams   []*audioStream
+	sync.Mutex
+	streams   map[string]*audioStream
 	expected  int
 	threshold uint8
 	previous  []string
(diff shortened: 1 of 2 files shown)


OrlandoCo:
besides reusing the string slice, which I agree why moving to map is more performant? Can we add a benchmark so we can be sure?

hn8:
@OrlandoCo For large session, current codepath need to sort all streams even quiet ones. But active speakers are usually very short 0-3 in practice.

OrlandoCo:
Even large sessions are small enough for a slice to get better performance than a map, we can make that the slice just get active speakers and on calc just clear it with [:0] until the next loop. We can benchmark both if required.

hn8:
@OrlandoCo Yap, in general, audio stream count is relatively small, not likely to have a few hundred audio publishers in practice. Do you know how often observe() is called? Is AudioLevelExtension included in every RTP audio packet? Currently, DataChannel in PeerLocal use map not slice, but DC count is also small per session. Lol, I feel embarassed for doing premature optimization while removing another one.

hn8:
@OrlandoCo I just implemented to use slice as you suggested, the problem of using slice  is that sort.Slice can't work on short streamIDs list (doesn't have a map to look up), so it is back to sorting full stream.
https://github.com/pion/ion-sfu/blob/d663b423c96b186e12458e61ba53e010d5ce092b/pkg/sfu/audioobserver.go#L82

To reuse previous slice, its type need to change from []string too. Do you wan to see the alternative version?

hn8:
@OrlandoCo Here is the alternative version using slice. IMHO, map version is simpler.
https://github.com/Screenleap/ion-sfu/commit/85d6678b9f1ec4b646406a2f0ffc213c15ee23ac

hn8:
@OrlandoCo Here is the alternative version using slice
https://github.com/Screenleap/ion-sfu/commit/154f24e8aeb1f0ea8c386b32b21bcafcd77861ae

PR #551, opened 2021-07-08
Use sequence parameter sets for h.264 keyframe detection

davidzhao:
Similar to the change made in Galene by jech, use SPS to detect keyframes. This fixes glitches in H.264 when layer switching

https://github.com/jech/galene/commit/4435a30a539a97c4dc2c062f19c6ff54b3930035


OrlandoCo (approved):
LGTM

PR #558, opened 2021-07-23
Add NoAutoSubscribe to peer.JoinConfig.

cloudwebrtc:
#### Description
Do not subscribe to any stream during peer join, so that the application layer can perform fine control and subscribe on-demand

1, Listen publisher.OnPublishTrack event, generate StreamEvent and broadcast to each peer.
```go
peer.Publisher().OnPublisherTrack(func(track ion_sfu.PublisherTrack) {
    log.Debugf("peer.OnPublisherTrack: \nKind %v, \nUid: %v,  \nMsid: %v,\nTrackID: %v",
                       track.Track.Kind(),
                       uid, 
                       track.Track.Msid(), 
                       track.Track.ID())
    // event := &rtc.StreamEvent{.....}
    // sig.BroadcastStreamEvent(event)
})
```

2,Peer uses the trackIds in StreamEvent to subscribe or unsubscribe.
```go
case *rtc.UpdateSettings_Subcription:

subscription := payload.UpdateSettings.GetSubcription()
enabled := subscription.GetSubscribe()

for _, trackId := range subscription.TrackIds {

	if enabled {
                 // Add down tracks
		for _, p := range peer.Session().Peers() {
			for _, track := range p.Publisher().PublisherTracks() {
				if track.Receiver.TrackID() == trackId {
					peer.Publisher().GetRouter().AddDownTrack(peer.Subscriber(), track.Receiver)
				}
			}
		}
	} else {
               // Remove down tracks
		for streamID, downTracks := range peer.Subscriber().DownTracks() {
			for _,  track := range downTracks {
				if track.ID() == trackId {
					peer.Subscriber().RemoveDownTrack(streamID, track)
                                          track.Stop()
				}
			}
		}
	}

}

peer.Subscriber().Negotiate()
```

### DIFF
--- a/pkg/sfu/publisher.go
+++ b/pkg/sfu/publisher.go
@@ -25,6 +25,7 @@ type Publisher struct {
 	candidates []webrtc.ICECandidateInit
 
 	onICEConnectionStateChangeHandler atomic.Value // func(webrtc.ICEConnectionState)
+	onPublisherTrack                  func(PublisherTrack)
 
 	closeOnce sync.Once
 }
@@ -72,10 +73,12 @@ func NewPublisher(id string, session Session, cfg *WebRTCTransportConfig) (*Publ
 		)
 
 		r, pub := p.router.AddReceiver(receiver, track)
+		var publisherTrack PublisherTrack
 		if pub {
 			p.session.Publish(p.router, r)
 			p.mu.Lock()
-			p.tracks = append(p.tracks, PublisherTrack{track, r, true})
+			publisherTrack = PublisherTrack{track, r, true}
+			p.tracks = append(p.tracks, publisherTrack)
 			for _, rp := range p.relayPeer {
 				if err = p.createRelayTrack(track, r, rp); err != nil {
 					Logger.V(1).Error(err, "Creating relay track.", "peer_id", p.id)
@@ -84,8 +87,13 @@ func NewPublisher(id string, session Session, cfg *WebRTCTransportConfig) (*Publ
 			p.mu.Unlock()
 		} else {
 			p.mu.Lock()
-			p.tracks = append(p.tracks, PublisherTrack{track, r, false})
+			publisherTrack = PublisherTrack{track, r, false}
+			p.tracks = append(p.tracks, publisherTrack)
 			p.mu.Unlock()
+		}
+
+		if p.onPublisherTrack != nil {
+			p.onPublisherTrack(publisherTrack)
 		}
 	})
 
@@ -159,6 +167,10 @@ func (p *Publisher) Close() {
 			Logger.Error(err, "webrtc transport close err")
 		}
 	})
+}
+
+func (p *Publisher) OnPublisherTrack(f func(track PublisherTrack)) {
+	p.onPublisherTrack = f
 }
 
 // OnICECandidate handler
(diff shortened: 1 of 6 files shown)


adwpc (approved):
LGTM. seems no break changes

pkg/sfu/publisher.go
```diff
@@ -84,9 +87,14 @@ func NewPublisher(id string, session Session, cfg *WebRTCTransportConfig) (*Publ
 			p.mu.Unlock()
 		} else {
 			p.mu.Lock()
-			p.tracks = append(p.tracks, publisherTracks{track, r, false})
+			publisherTrack = PublisherTrack{track, r, false}
+			p.tracks = append(p.tracks, publisherTrack)
 			p.mu.Unlock()
 		}
+
+		if p.onPublisherTrack != nil {
```
OrlandoCo, line 95: you should call the method only when pub is true otherwise on simulcast you will fire the event up to 3 times.

pkg/sfu/publisher.go
```diff
@@ -20,22 +20,23 @@ type Publisher struct {
...
-	track    *webrtc.TrackRemote
-	receiver Receiver
+type PublisherTrack struct {
+	Track    *webrtc.TrackRemote
+	Receiver Receiver
 	// This will be used in the future for tracks that will be relayed as clients or servers
 	// This is for SVC and Simulcast where you will be able to chose if the relayed peer just
 	// want a single track (for recording/ processing) or get all the tracks (for load balancing)
-	clientRelay bool
+	ClientRelay bool
```
OrlandoCo, line 39: no need to make this field public, is still experimentatl and private

pkg/sfu/publisher.go
```diff
@@ -72,10 +73,12 @@ func NewPublisher(id string, session Session, cfg *WebRTCTransportConfig) (*Publ
 		)
 
 		r, pub := p.router.AddReceiver(receiver, track)
+		var publisherTrack PublisherTrack
 		if pub {
 			p.session.Publish(p.router, r)
 			p.mu.Lock()
-			p.tracks = append(p.tracks, publisherTracks{track, r, true})
+			publisherTrack = PublisherTrack{track, r, true}
```
OrlandoCo, line 80: if PublisherTrack is already declared no need to re asign

pkg/sfu/publisher.go
```diff
@@ -161,6 +169,10 @@ func (p *Publisher) Close() {
 	})
 }
 
+func (p *Publisher) OnPublisherTrack(f func(track PublisherTrack)) {
+	p.onPublisherTrack = f
```
OrlandoCo, line 173: should we use atomics here?

pkg/sfu/subscriber.go
```diff
@@ -215,6 +218,16 @@ func (s *Subscriber) GetDatachannel(label string) *webrtc.DataChannel {
 	return s.DataChannel(label)
 }
 
+func (s *Subscriber) DownTracks() map[string][]*DownTrack {
```
OrlandoCo, line 221: This would be better as an slice, maps are expensive when `n` is small

cloudwebrtc:
@OrlandoCo updated

PR #561, opened 2021-08-03
feat(relay): Add relay ingest to SFU

OrlandoCo:
#### Description

This PR brings lots of improvements to ION:

1. Allows to ingest relayed streams.
2. Lock free RTP writer
3. Sender reports time propagation
4. And some other minor fixes

There still some things to consider while using relay

- [ ] No simulcast supported (WIP), atm simulcast is handled as a normal peer so it relays the highest or lower layer according conf
- [ ] Audioobserver not supported on relaying SFU
- [ ] Datachannel fanouts are supported but not well tested yet


adwpc (approved):
cool！ortc and atomic make things simple

pkg/sfu/receiver.go
```diff
@@ -360,52 +334,80 @@ func (w *WebRTCReceiver) writeRTP(layer int) {
 			return
 		}
 
-		w.locks[layer].Lock()
+		if w.isSimulcast {
+			pts := w.pendingTracks[layer].Load().([]*DownTrack)
+			if len(pts) > 0 {
+				if pkt.KeyFrame {
+					w.Lock()
+					for _, dt := range pts {
```
davidzhao, line 342: I think we'll need to reload pts here in order to avoid the race.

pkg/sfu/receiver.go
```diff
@@ -360,52 +334,80 @@ func (w *WebRTCReceiver) writeRTP(layer int) {
...
 			}
 		}
 
-		for _, dt := range w.downTracks[layer] {
-			if err = dt.WriteRTP(pkt); err != nil {
+		dts := w.downTracks[layer].Load().([]*DownTrack)
+		for _, dt := range dts {
+			if err = dt.WriteRTP(pkt, layer); err != nil {
+				if err == io.EOF && err == io.ErrClosedPipe {
+					w.Lock()
```
davidzhao, line 360: I think we perform this in several places, Lock then DeleteDownTrack. should the lock/unlock be moved into DeleteDownTrack?

pkg/sfu/receiver.go
```diff
@@ -360,52 +334,80 @@ func (w *WebRTCReceiver) writeRTP(layer int) {
...
+func (w *WebRTCReceiver) storeDownTrack(layer int, dt *DownTrack) {
+	dts := w.downTracks[layer].Load().([]*DownTrack)
+	ndts := make([]*DownTrack, len(dts)+1)
+	copy(ndts, dts)
+	ndts[len(ndts)-1] = dt
+	w.downTracks[layer].Store(ndts)
+}
+
+func (w *WebRTCReceiver) storePendingTrack(layer int, dt *DownTrack) {
+	w.Lock()
```
davidzhao, line 406: what's the typical convention that's used here? for where the locks should be? just wondering why `storeDownTrack` expects callers to lock, while `storePendingTrack` locks within the function.

davidzhao (commented):
this is some pretty exciting stuff!

pkg/sfu/receiver.go
```diff
@@ -257,32 +236,22 @@ func (w *WebRTCReceiver) DeleteDownTrack(layer int, id string) {
 		return
 	}
 
-	w.locks[layer].Lock()
-	idx := -1
-	for i, dt := range w.downTracks[layer] {
-		if dt.peerID == id {
-			idx = i
-			break
+	dts := w.downTracks[layer].Load().([]*DownTrack)
```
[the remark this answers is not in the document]
davidzhao: nvm, I see you the caller of DeleteDownTrack is locking.

pkg/sfu/receiver.go
OrlandoCo: Originally was like that, but on layer change we want to perform many operations, that means that we would need to get a lock and free it 3 times.

pkg/sfu/receiver.go
OrlandoCo: Since pending track are stored only in one function I put the lock there, but store to downtrack are called in other places, so to not lock many times for example:
```go
					for _, dt := range pts {
						w.DeleteDownTrack(dt.CurrentSpatialLayer(), dt.peerID)
						w.storeDownTrack(layer, dt)
						dt.SwitchSpatialLayerDone(int32(layer))

					}
```
Above code would require to get the locks many times.

pkg/sfu/receiver.go
OrlandoCo: Agree there is super small probability that during the first Load and get the lock something modifies the slice.

pkg/sfu/receiver.go
```diff
@@ -118,56 +119,43 @@ func (w *WebRTCReceiver) AddUpTrack(track *webrtc.TrackRemote, buff *buffer.Buff
...
+	w.downTracks[layer].Store(make([]*DownTrack, 0, 10))
+	w.pendingTracks[layer].Store(make([]*DownTrack, 0, 10))
 	w.Unlock()
 
 	subBestQuality := func(targetLayer int) {
 		for l := 0; l < targetLayer; l++ {
-			w.locks[l].Lock()
-			for _, dt := range w.downTracks[l] {
-				dt.SwitchSpatialLayer(int64(targetLayer), false)
+			dts := w.downTracks[l].Load()
```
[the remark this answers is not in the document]
OrlandoCo: Slice modifiers ara always locked, only getters are atomics

pkg/sfu/downtrack.go
```diff
@@ -345,9 +341,10 @@ func (d *DownTrack) writeSimpleRTP(extPkt *buffer.ExtPacket) error {
 			}
 		}
 
-		d.snOffset = extPkt.Packet.SequenceNumber - d.lastSN - 1
-		d.tsOffset = extPkt.Packet.Timestamp - d.lastTS - 1
-
+		if d.lastSN != 0 {
```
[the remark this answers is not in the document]
OrlandoCo: with sync we always set timestamps to 1 and increment from there according to the offsets, this will just set the first timestamp the same as the source and just make sure that we follow spec which says that TS and SN should be random.

pkg/sfu/receiver.go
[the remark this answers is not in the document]
davidzhao: right. there isn't any data loss, though a newly added layer might not be placed onto the desired layer.. it should automatically correct itself over time.

pkg/sfu/receiver.go
davidzhao: that makes sense. I think functionality-wise that's optimal. I do wonder if there's a way to make this more obvious to the reader. I've seen others use the convention of public/private functions where locking is handled in the public.

OrlandoCo: Yeah I was thinking on doing that for `DeleteDownTrack` other publics methods are correctly protected

pkg/relay/README.md
```diff
@@ -0,0 +1,81 @@
...
+   if meta.session== "world-domination"{
+      	return RelayToLegionOfDoom(meta, signal)
+   }	
+   return nil, errors.New("not supported")
+ }
+ 
+ // The remote peer should create a new Relay Peer with the metadata and call Answer.
+ if err:= peer.Offer(signalFunc); err!=nil{
+ 	handleErr(err)
+ }
```
tab1293, line 37: This above usage is wrapped up in [Publisher's Relay](https://github.com/pion/ion-sfu/blob/feat/relayingest/pkg/sfu/publisher.go#L195) method correct?

tab1293:
Great work! This is an awesome feature addition 🎉

pkg/relay/README.md
```diff
@@ -0,0 +1,81 @@
...
+ // you should fill according your requirements or leave the defaults.
+ config := &PeerConfig{} 
+ peer, err := NewPeer(meta, config)
+ handleErr(err)
+ 
+ // Now before working with the peer you need to signal the peer to 
+ // your remote sever, the signaling can be whatever method you want (gRPC, RESt, pubsub, etc..)
+ signalFunc= func (meta PeerMeta, signal []byte) ([]byte, error){
+   if meta.session== "world-domination"{
+      	return RelayToLegionOfDoom(meta, signal)
```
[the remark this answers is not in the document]
OrlandoCo: Is correct at the bottom of the readme there is information on how to use it with ion-sfu to ion-sfu, so you just need to relay the byte slice and consume it on `session.AddRelayPeer` as you said, the returned byte slice from `AddRelayPeer` would be the answer to signal func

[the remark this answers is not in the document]
tab1293: Ah I missed that bottom section 😬 This all makes sense and is a convenient API, thanks for this documentation

go.mod
```diff
@@ -8,27 +8,31 @@ require (
...
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/improbable-eng/grpc-web v0.13.0
 	github.com/lucsky/cuid v1.0.2
 	github.com/pion/dtls/v2 v2.0.9
-	github.com/pion/ice/v2 v2.1.8
+	github.com/pion/ice/v2 v2.1.10
 	github.com/pion/logging v0.2.2
 	github.com/pion/rtcp v1.2.6
-	github.com/pion/rtp/v2 v2.0.0
+	github.com/pion/rtp v1.6.5
```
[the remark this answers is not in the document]
OrlandoCo: V2 is a breaking change that is not going to be merged until v4 is ready or something else happens. Is there any reason you need v2?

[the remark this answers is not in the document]
adwpc: No. not nessary

yifan513:
Is there any example of how to use the relay just to connect two ion-sfu?

adwpc:
> Is there any example of how to use the relay just to connect two ion-sfu?

A simple document is here
https://github.com/pion/ion-sfu/tree/master/pkg/relay

PR #573, opened 2021-08-22
Add notification to the middleware of peer removal.

### LINKED ISSUE #572: Datachannel middleware notification of Peer dropping.
## Summary

When a peer drops the middleware is not notified. This information could be useful to some middlewares.

## Motivation

A middleware should be free to do complex things depending on the peers that actively connected. When a peer drops the middleware should get notified so that it may adjust its internal state.

tmiv:
#### Description

This change allows data channel middleware to register a callback for when a peer disconnects.

#### Reference issue
Fixes #572

### DIFF
--- a/pkg/sfu/datachannel.go
+++ b/pkg/sfu/datachannel.go
@@ -14,6 +14,7 @@ type (
 		Label       string
 		middlewares []func(MessageProcessor) MessageProcessor
 		onMessage   func(ctx context.Context, args ProcessArgs)
+		onRemovePeer func( peer Peer )
 	}
 
 	ProcessArgs struct {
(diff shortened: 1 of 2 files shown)


OrlandoCo:
Hey @tmiv  sorry for late response, but is necessary to put the callback on the datachannel?, should `Peer` be a better location for a close callback?

tmiv:
No @OrlandoCo, I don't think so. It's the middleware who should process the disconnect (onRemovePeer).  Just as the middleware processes the messages (onMessage).  But I think that I might be convoluting the concept of a dc middleware here. It seems that data channel can have more than one middleware thus the DC chain structure. Also, that the concept of a middleware is really just a single function.

PR #590, opened 2021-09-25
refactor(logger): switch to go-logr/zerologr

hn8:
Description

Use optimized implementation instead of json.Marshal in e.Interface()


hn8:
@OrlandoCo @tarrencev @adwpc Would you please review simplyfing custom code in logger? After merging PRs in upstream rs/zerolog and logr, this change will switch to the official implementation hosted by go-logr instead. Also reduce memory allocation from e.Interface(). Thanks

adwpc (approved):
LGTM

PR #591, opened 2021-09-30
fix(buffer): VP8 should not fail early on packet len

OrlandoCo:


AnshulMalik:
@OrlandoCo what's the reason for this change?

PR #598, opened 2021-10-14
Fix prometheus statistics

jongwow:
#### Description
- sessions prometheus stats not working
- Peers not registered


jongwow:
invalid author

PR #608, opened 2021-11-09
add media-device broadcast example

mohammadne:
HI @Sean-Der 

#### Description
add `media-device broadcast` example which `Go` peer will publish camera of itself
and the other `JS` peer will consume the media


mohammadne:
@Sean-Der 
any update ?

adwpc:
@mohammadne 
`# github.com/pion/mediadevices
../../../../../../pkg/mod/github.com/pion/mediadevices@v0.2.0/track.go:342:33: cannot use mtu (type int) as type uint16 in argument to rtp.NewPacketizer
../../../../../../pkg/mod/github.com/pion/mediadevices@v0.2.0/track.go:462:33: cannot use mtu (type int) as type uint16 in argument to rtp.NewPacketizer`
i got this error when use mediadevices@v0.2.0
it worked when i change to mediadevices@v0.3.0
there is no audio

mohammadne:
Hi @adwpc 
I have just updated the `mediadevices`

adwpc:
@mohammadne 
thanks, but i think it's better not to add more examples in ion-sfu/example.
and better to use ion-sdk-go to publish from devices
https://github.com/pion/ion-sdk-go/tree/master/example

mohammadne:
@adwpc  ok, no problem
but in overall I think having the same functionality in both sdk and library can help alot.
thanks for your attention

PR #622, opened 2021-11-29
Add other end to end projects that use ion-sfu

davidzhao:
Adding end to end solutions livekit/ion to guide users looking for such.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,21 @@
 A [selective forwarding unit](https://webrtcglossary.com/sfu/) is a video routing service which allows webrtc sessions to scale more efficiently. This package provides a simple, flexible, high performance Go implementation of a WebRTC SFU. It can be called directly or through a [gRPC](cmd/signal/grpc) or [json-rpc](cmd/signal/json-rpc) interface.
 
 ## Features
+
 * Audio/Video/Datachannel forwarding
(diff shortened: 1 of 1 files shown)


OrlandoCo (approved):

adwpc:
@davidzhao @OrlandoCo 
better not to add projects entrance here because there are so many projects using ion-sfu.
That will make readme unmaintainable if all the projects want add a entrance here

OrlandoCo:
@adwpc I think is useful for non experts on webrtc to have a batteries included projects referenced on the read me, people often gets confused by the complexity of ion-sfu, or try to add more functionality that are out of scope of the SFU. This entries should not be for all projects who uses ion-sfu, but only for distributed conference systems like ION and Livekit which can help users to quick bootstrap their own projects without getting to deep into the webrtc stack.

davidzhao:
@adwpc I can understand your concern. I think we ought to put in a clear statement of what (and whom) ion-sfu is targeted at. Too often do we see folks not having the right set of expectations coming into slack; thinking it to be a full conferencing solution.

This may be self-serving, but I think we are helping folks by pointing them towards proven OSS projects that do provide end-to-end solutions; versus listing everything that uses ion-sfu.

PR #627, opened 2021-12-07
[FIX]  Datachannle already exists

zjzhang-cn:
#### Description
[FIX] When DataChannel Label already exists , The message of the new DC cannot be forwarded.
#### Reference issue
Fixes #...


adwpc (approved):
Thanks！

PR #631, opened 2021-12-14
FIX: The creation of multiple relay datachannel failed. Procedure

upeartaker:
Hello my friend,
The ID in DataChannelParameters should be incremented, there are some problems with the current way of writing, it is 1 every time, I dealt with this problem, you can review it if you can.

At the same time, I found that relay does not process the Middleware Datachannel at present, so some functions implemented by ion-sfu datachannel cannot run properly under relay. I fixed this problem, and we can discuss it if necessary.


kevmo314 (approved):
Looks good, thanks for the fix!

PR #637, opened 2022-01-13
Fix crash when audio nack packets received

lenaky:
#### Description

If client sent rtcp.TransportLayerNack to audio downtrack,
ion-sfu going crash because downTrack.sequencer is only allocated for video track.
Check sequencer is nil when rtcp.TransportLayerNack has arrived.

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/sfu/downtrack.go
+++ b/pkg/sfu/downtrack.go
@@ -532,12 +532,14 @@ func (d *DownTrack) handleRTCP(bytes []byte) {
 				}
 			}
 		case *rtcp.TransportLayerNack:
-			var nackedPackets []packetMeta
-			for _, pair := range p.Nacks {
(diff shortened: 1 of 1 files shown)


OrlandoCo:
well is more weird that you are receiving a nack packet for audio,  this is from a browser client?

lenaky:
@OrlandoCo yes, we've received from `safari 15.2, mac 10.15.6`. 
Anyway I think It is more safer checking sequencer is nil because we don't know how client act.

OrlandoCo (approved):

Sean-Der:
@OrlandoCo ok if I merge this?

OrlandoCo:
@Sean-Der Merged!

PR #642, opened 2022-01-24
Fix(receiver): Fix that Close method is not called when downtrack is deleted.

YMLee3218:
#### Description

onCloseHandler of downtrack was not called because close method of downtrack was called when receiver was closed and not when only downtrack was deleted.


adwpc (approved):
LGTM

PR #687, opened 2023-07-21
fix: simulcast vp8

atdevten:
Avoid layer deleted in change.
Remove feedback (client calculated outgoing bitrate). it makes the vp8 simulcast freeze at high quality.


AnshulMalik:
@ATtendev what's the reasoning behind removal of rtcp feedback? It sends reception reports as well.

